PR #61, opened 2022-03-11
changed final purity above 1 to equal 1 and fix some bugs in purity e…

kristenadaley:
…stimation

# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1113,7 +1113,7 @@ estimate_purity = function(in_maf,
   }else{
     # If no seg file was provided and assume_diploid paramtere is set to true,
     if(assume_diploid){
-      CN_new <- assign_cn_to_ssm(maf_file = in_maf, assume_diploid = TRUE,coding_only=coding_only,genes=genes)
+      CN_new <- assign_cn_to_ssm(maf_file = in_maf, assume_diploid = TRUE,coding_only=coding_only,genes=genes)$maf
     }
   }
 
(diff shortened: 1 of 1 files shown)


R/utilities.R
```diff
@@ -1157,10 +1157,11 @@ estimate_purity = function(in_maf,
   # Calculate purity, and if the number is larger 1 (100%) use the VAF calue instead
```
mattssca, line 1157: Found a tiny typo, "VAF value", not "calue". But this was something previous.

R/utilities.R
```diff
@@ -1113,7 +1113,7 @@ estimate_purity = function(in_maf,
   }else{
     # If no seg file was provided and assume_diploid paramtere is set to true,
     if(assume_diploid){
-      CN_new <- assign_cn_to_ssm(maf_file = in_maf, assume_diploid = TRUE,coding_only=coding_only,genes=genes)
+      CN_new <- assign_cn_to_ssm(maf_file = in_maf, assume_diploid = TRUE,coding_only=coding_only,genes=genes)$maf
```
mattssca, line 1116: Is the seq list redundant (i.e not needed for downstream code), or what's the reason for dropping the seq list in this PR?

mattssca (approved):
Nice edits, I've tried the updated function and it does what's advertised. No changes requested, see comments in PR for more info.

R/utilities.R
kristenadaley: I am keeping the list but just have to specify the maf (at the end), otherwise it can throw an error

PR #73, opened 2022-04-08
New functions: calculate PGA and adjust ploidy

Kdreval:
This PR addresses a bug that broke column ordering in get_gambl_metadata, but also it adds new functionality to calculate the proportion of genome altered (PGA) by CNV and to adjust sample ploidy based on the PGA. The new functions work in both projections (grch37 and hg38), operate on single- or multi-sample files, and have an option to read the file from a local path if necessary.


R/database.R
```diff
@@ -597,9 +597,9 @@ get_gambl_metadata = function(seq_type_filter = "genome",
 
   all_meta = GAMBLR::collate_lymphgen(all_meta, verbose=FALSE)
 
-# "catchall" pathology for those that need review
-all_meta = all_meta %>%
-  mutate(pathology = ifelse(nchar(pathology) > 15, "OTHER", pathology))
+  # "catchall" pathology for those that need review
+  all_meta = all_meta %>%
+    mutate(pathology = ifelse(nchar(pathology) > 15, "OTHER", pathology))
```
rdmorin, line 602: Haha. That's a good way to deal with it!

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+#' sample_seg = get_sample_cn_segments(this_sample_id = "14-36022T") %>% rename("sample"="ID")
+#' calculate_pga(this_seg = sample_seg)
+#' calculate_pga(this_seg = sample_seg, exclude_sex = FALSE)
+#'
+#' multi_sample_seg = rbind(get_sample_cn_segments(this_sample_id = "14-36022T"),
+#'                          get_sample_cn_segments(this_sample_id = "BLGSP-71-21-00243-01A-11E")) %>%
+#'                          rename("sample"="ID")
+#' GAMBLR::calculate_pga(this_seg = multi_sample_seg)
+#'
+
```
rdmorin, line 2678: Didnt you have some package you were using earlier in GAMBLR that did this calculation for you?

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+#' This function calculates the percent of genome altered (PGA) by CNV. It takes into account the total length of
+#' sample's CNV and relates it to the total genome length to return the proportion affected by CNV. The input is expected to be seg file.
+#' The path to a local SEG file can be provided instead. If The custom seg file is provided, the minimum required columns are
+#' sample, chrom, start, end, and log.ratio. The function can work with either individual or multi-sample seg file. The telomeres are always
+#' excluded from calculation, and centromeres/sex chromosomes can be optionally included or excluded.
+#'
+#' @param this_seg Input data frame of seg file.
+#' @param seg_path Optionally, specify the path to a local seg file.
+#' @param projection Argument specifying the projection of seg file, which will determine chr prefix, chromosome coordinates, and genome size. Default is grch37, but hg38 is also accepted.
+#' @param cutoff The minimum log.ratio for the segment to be considered as CNV. Default is 0.56, which is 1 copy. This value is expected to be positive float of log.ratio for both deletions and amplifications.
```
rdmorin, line 2660: Does this mean only gains are considered in PGA?

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+
+  # now add any samples that can have 0 PGA
+  affected_regions = base::merge(sample_set,
+                                 affected_regions,
+                                 all.x = TRUE) %>%
+    replace_na(list(PGA = 0))
+
+  affected_regions = affected_regions %>%
+    select(-total) %>%
+    `names<-`(c("sample_id", "PGA")) %>%
```
rdmorin, line 2786: How does this work?

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+  # exclude sex chromosomes
+  if (exclude_sex) {
+    this_seg = this_seg %>%
+      dplyr::filter(!grepl("X|Y", chrom))
+  }
+
+  # if PGA is called with custom parameters or obtained elsewhere, use it - but if not, calculate on-the-fly
+  if (missing(pga)) {
+    message("Calculating PGA ...")
+    pga = calculate_pga(this_seg = this_seg,
```
rdmorin, line 2858: I think pga needs to be calculated per sample in multi-sample segs but it doesn't seem like this is being done or is pga a vector of many pgas from a multi-sample seg?

rdmorin: Never mind. I answered my own question. Nicely done!

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+    )
+  }
+
+  # if the seg is a local file, read it in
+  if (!missing(seg_path)) {
+    message(paste0("Reading thhe seg file from ", seg_path))
+    this_seg = read_tsv(seg_path)
+  }
+
+  # ensure consistent chromosome prefixing
```
rdmorin, line 2841: Does this require seg files to have specific headers? I thought some of our headers differed in the naming of the columns but I could be wrong.

R/utilities.R
```diff
@@ -2646,3 +2646,267 @@ standardize_chr_prefix = function(incoming_vector,
...
+  # does the user's seg file contain centromeres?
+  if (exclude_centromeres) {
+    chr_coordinates = chr_coordinates %>%
+      group_by(chromosome) %>%
+      mutate(start = min(start),
+             end = max(end)) %>%
+      ungroup
+  }
+
+  # total size of genome in this projection
```
rdmorin, line 2716: Suggestion: put this code outside any function such that it runs when GAMBLR is loaded and sets a global variable that can be used from anywhere:

genome_sizes = list(grch37=genome_size_grch37, hg38=genome_size_hg38)

inst/extdata/chromosome_arms_grch37.tsv
```diff
@@ -0,0 +1,49 @@
+chromosome	start	end	arm
+1	10000	121500000	p
+1	142600000	249250621	q
```
rdmorin, line 3: Handy! I was getting these from a package for another function but can switch that code to use this directly

rdmorin (approved):
No changes necessary but some you may want to consider

R/utilities.R
Kdreval: This makes sense, thanks! I'll do it in the next update and make this global for GAMBLR

R/utilities.R
Kdreval: I specify the function within the tickmarks and then in parentheses what should go there. It can also recognize other functions like `rownames` etc - a good approach to renaming columns/rows in pipe. So here I just specified the column names

PR #82, opened 2022-04-11
fix flatfile reading in get_manta_sv

rdmorin:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


Kdreval (approved):
Thanks!

R/annotation.R
```diff
@@ -219,15 +219,15 @@ annotate_driver_ssm = function(maf_df,
...
   }else{
     #this all assumes the user has left the noncoding variants in their MAF
-    kept_ssm = maf_df %>% 
+    kept_ssm = maf_df %>%
       dplyr::filter(Variant_Classification %in% coding_vc) %>%
       dplyr::filter(Hugo_Symbol %in% driver_genes)
 
     for(gene in names(include_noncoding)){
       message(paste("adding", unname(include_noncoding[gene]), "for", gene))
-      nc_ssm = maf_df %>%
```
mattssca, line 228: Are all of these edits related to the feature you told us about last week, where Rstudio automatically cleans up white stray spaces etc? Handy!

R/annotation.R
```diff
@@ -266,20 +266,20 @@ annotate_driver_ssm = function(maf_df,
...
-                       partner_bed, 
-                       with_chr_prefix = FALSE, 
-                       collapse_redundant = FALSE, 
+annotate_sv = function(sv_data,
+                       partner_bed,
+                       with_chr_prefix = FALSE,
+                       collapse_redundant = FALSE,
                        return_as = "bedpe_entrez",
-                       blacklist = c(60565248, 30303126, 187728894, 101357565, 101359747, 161734970, 69400840, 65217851, 187728889, 188305164),
+                       blacklist = c(60565248, 30303126, 187728894, 101357565, 101359747, 161734970, 69400840, 65217851, 187728889, 187728888,187728892, 187728893,188305164),
```
mattssca, line 274: Should this blacklist parameter I implemented be pointing to a bed file instead? Or is it enough, for now, to just state such regions like this?

mattssca (approved):
Thanks for the updates, Ryan. Do you want to regenerate the documentation to have it reflect the added (and updated) parameters or this can wait until the remaining issues are resolved?

R/annotation.R
rdmorin: We should definitely move this information to a bundled file

PR #84, opened 2022-04-13
Bug fix: positional arguments in get_ssm_by_samples call to get_ssm_b…

lkhilton:
This small patch names the arguments in the call to `get_ssm_by_sample` in `get_ssm_by_samples`. Previously the `flavour` argument was in the `augmented` position and so logical tests in `get_ssm_by_sample` were erroring out. 

I see this PR is affected by the ongoing whitespace issue and many of these changes have to do only with RStudio trimming the whitespace. The relevant lines are 144-152.


mattssca (approved):
Thanks Laura!

Issue #19, opened 2021-06-26
some segmented data not found for unmatched tumours


rdmorin:
Some of the controlfreec results seem to be missing from GAMBLR. It might be that hg38 lifted segmented data from controlfreec were never loaded into the database. Is this a new thing?

```
all_segs_unmatched %>% pull(ID) %>% unique()
  [1] "00-14595_tumorC"           "00-15201_tumorA"           "00-15201_tumorB"          
  [4] "00-23442_tumorA"           "00-23442_tumorB"           "00-26427_tumorA"          
  [7] "01-12047_tumorB"           "01-12047_tumorC"           "01-14774_tumorA"          
 [10] "01-14774_tumorB"           "01-16433_tumorA"           "01-16433_tumorB"          
 [13] "01-20774T"                 "01-23117_tumorA"           "01-23117_tumorB"          
 [16] "01-23942_tumorA"           "01-23942_tumorB"           "01-28152_tumorA"          
 [19] "01-28152_tumorB"           "02-13135T"                 "02-15630_tumorA"          
 [22] "02-15630_tumorB"           "02-18356_tumorA"           "02-18356_tumorB"          
 [25] "02-20170T"                 "02-22991T"                 "02-24492_tumorA"          
 [28] "02-28397_tumorA"           "02-28397_tumorB"           "03-10363T"                
 [31] "03-10440_tumorA"           "03-10440_tumorB"           "03-11110T"                
 [34] "03-19103_tumorB"           "03-23488_tumorA"           "03-23488_tumorB"          
 [37] "03-33266_tumorA"           "03-33266_tumorB"           "04-14066_tumorA"          
 [40] "04-14066_tumorB"           "04-14093_tumorA"           "04-14093_tumorB"          
 [43] "04-21856_tumorA"           "04-24061_tumorA"           "04-24061_tumorB"          
 [46] "04-24937T"                 "04-28140T"                 "04-35039T"                
 [49] "05-12939T"                 "05-17793T"                 "05-18426T"                
 [52] "05-21634T"                 "05-22052T"                 "05-23110T"                
 [55] "05-24065T"                 "05-24395T"                 "05-24401T"                
 [58] "05-24666T"                 "05-25439T"                 "05-25674T"                
 [61] "05-32150T"                 "05-32150_tumorA"           "05-32150_tumorB"          
 [64] "05-32762T"                 "05-32947T"                 "06-10398T"                
[...]

PR #86, opened 2022-04-25
Implement flatfile functionality for CNV and introduce cnvKompare

Kdreval:
This PR introduces the flatfile functionality for CNV to avoid reliance on the database. The useful feature here is a new `cnvKompare` function that will compare CNV data for patients with multiple time points and calculate the overall concordance of CNV + other comparison data. It can also handle the comparison of CNV between different tools if needed.


Kdreval:
This also addresses and resolves issue #19

R/utilities.R
```diff
@@ -2910,3 +2910,310 @@ adjust_ploidy = function(this_seg,
 
   return(this_seg)
 }
+
+
+#' This function will compare CNV data between samples with multiple time points. It can also handle same-sample comparison
```
lkhilton, line 2915: This format will put this entire block of text into the title so when you look at the help for this function all these details are in large bold font at the top. The format should go like this: 

```
#' Sum of vector elements
#'
#' `sum` returns the sum of all the values present in its arguments.
#'
#' This is a generic function: methods can be defined for it directly
#' or via the [Summary()] group generic. For this to work properly,
#' the arguments `...` should be unnamed, and dispatch is on the
#' first argument.
```

(From https://cran.r-project.org/web/packages/roxygen2/vignettes/rd.html)

Perhaps while you fix this you could update the docstrings for `sanitize_maf_data`?

lkhilton (changes requested):
Awesome! Just needs an update to the docstrings.

R/utilities.R
```diff
@@ -2910,3 +2910,310 @@ adjust_ploidy = function(this_seg,
...
+
+  # check that sample identifiers are provided
+  if (missing(patient_id) & missing(sample_ids)) {
+    stop("Please provide patient id or sample ids for comparison.")
+  }
+
+  # retrieve sample ids if only patient id is specified
+  if (missing(sample_ids)) {
+    sample_ids = get_gambl_metadata()
+    sample_ids = dplyr::filter(sample_ids, patient_id == {
```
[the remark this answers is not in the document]
Kdreval: Ah, this is syntax autoformatting! Have to be careful not to miss this the next time!

R/utilities.R
Kdreval: Thanks, I'll fix this!

Kdreval:
Thanks for review Laura! I fixed these issues, and also added additional function to specify minimum similarity between cytobands to be considered concordant/discordant

lkhilton (approved):
Looks great! Adding percent concordance as a configurable parameter is a great addition.

PR #88, opened 2022-05-06
Bug fixes, new plotting functions, vignette examples, etc.

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

- [x] My function uses a library that isn't already a dependency of GAMBLR and I made the package aware of this dependency using the function documentation `import` statment. 

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


mattssca:
Updates in this PR address resolving existing bugs, quality-of-life updates and the addition of new plotting functions for SVs. The updates include:

1. get_coding_ssm_status has been updated to call get_coding_ssm to reduce duplicated code.

2. get_coding_ssm have been updated from seq_type to seq_type_filter (previously returned an error: Error in get_gambl_metadata(from_flatfile = from_flatfile, seq_type = seq_type) : argument 2 matches multiple formal arguments).

3. In fixing the above-mentioned issue, pretty_lollipop_plot in the vignette is now working the intended way again (issue closed).

4. prettyOncoplot in the vignette has been fixed, the issue has been closed.

5. get_coding_ssm with flat_file set to TRUE incorrectly caused the function to try to read from a maf path (non-existing since db is used to retrieve data).

6. get_ssm_by_region has been updated to make use of recent indexed (tabix) MAFs with respect to selected genome projection. Thus, the chunk that deals with regions has been updated to deal with chr prefixes accordingly. 

7. Flipping colours of SV counts in fancy_ideogram (previously incorrectly labelled deletions as indels and the other way around).

8. New parameters added to a collection of fancy_x_plots to allow the user to plot an already loaded (MAF-like) df in R, or directly specify a path to such a file. This allows for more flexibility to the affected plotting functions where the user can run these functions in an interactive session without having to run assign_cn_states or get_sample_cn_segments to retrieve plotting data. 

9. New plotting function added - fancy_circos_plot. This function imports RCircos and constructs a highly customizable circos plot. Per default, this plot visualizes translocations (ribbons). Optional arguments will create additional tracks to plot SVs. The user can also annotate the circos plot with gene lists. Optional parameters for VAF filtering, chromosome subsetting and genomic projection is also available. This function calls get_combined_sv to retain SV information (deletions, duplications and translocations).

10. New plotting function added - fancy_sv_size_plot, Generate plot visualizing SV sizes retreived with get_combined_sv. Subset on variant type, filter on VAF, size etc. 

11. GAMBLR documentation has been updated to reflect the current state and changes included in this PR. All new functions, changes to existing parameters, examples, NAMESPACE etc. have been documented adequately.

R/database.R
```diff
@@ -1505,23 +1504,24 @@ get_ssm_by_region = function(chromosome,
...
     #test if we have permissions for the full gambl + icgc merge
     if(mode == "slms-3"){
-      if(allow_clustered && !augmented){
+      if(!augmented){
         maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$deblacklisted
       }
-      if(allow_clustered && augmented){
+      if(augmented){
         maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$augmented
       }
```
rdmorin, line 1523: Why isn't this the more logical:

if(augmented){
maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$augmented
}else{
maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$deblacklisted
}

R/database.R
```diff
@@ -1530,23 +1530,29 @@ get_ssm_by_region = function(chromosome,
...
+  region = "chr8:128,723,128-128,774,067"
+
   if(!region == ""){
     region = gsub(",", "", region)
-    #format is chr6:37060224-37151701
     split_chunks = unlist(strsplit(region, ":"))
-    region = stringr::str_replace(region, "chr", "")
+    if(projection == "grch37"){
+      region = stringr::str_replace(region, "chr", "")
+    }
```
rdmorin, line 1546: What endless fun!

R/database.R
```diff
@@ -1648,9 +1654,9 @@ get_coding_ssm = function(limit_cohort,
   }else{
     all_meta = get_gambl_metadata(from_flatfile = from_flatfile, seq_type_filter = seq_type)
   }
-  all_meta = dplyr::filter(all_meta,seq_type == {{ seq_type }})
 
-  all_meta = get_gambl_metadata(from_flatfile = from_flatfile, seq_type = seq_type)
+  all_meta = dplyr::filter(all_meta,seq_type == {{ seq_type }})
+  all_meta = get_gambl_metadata(from_flatfile = from_flatfile, seq_type_filter = seq_type)
```
rdmorin, line 1659: Should this line be removed? It seems like these two lines are attempting to accomplish the same thing different ways.

R/utilities.R
```diff
@@ -129,46 +129,27 @@ get_coding_ssm_status = function(gene_symbols,
 
   if(missing(gene_symbols)){
     message("defaulting to all lymphoma genes")
-    gene_symbols = pull(lymphoma_genes,Gene)
```
rdmorin, line 132: Definitely makes sense to require the metadata as an argument here to avoid calling this many times.

R/utilities.R
```diff
@@ -129,46 +129,27 @@ get_coding_ssm_status = function(gene_symbols,
...
-  }else if (from_flatfile && augmented){
-    base_path = config::get("project_base")
-    maf_partial_path = config::get("results_flatfiles")$ssm$template$cds$augmented
-    maf_path = glue::glue(maf_partial_path)
-    full_maf_path = paste0(base_path, maf_path)
-    message(paste("reading from:", full_maf_path))
-    maf_data = fread_maf(full_maf_path)
-    coding_ssm = maf_data %>%
-      dplyr::filter(Variant_Classification %in% coding_class) %>%
-      dplyr::filter(t_alt_count >= min_read_support)
```
rdmorin, line 169: Nice change to make this code more concise!

R/viz.R
```diff
@@ -2742,16 +2746,34 @@ splendidHeatmap = function(this_matrix,
...
+    maf = fread_maf(maf_path)
+    maf = as.data.frame(maf)
+    colnames(maf)[variant_type_col] = "Variant_Type"
+    colnames(maf)[chromosome_col] = "Chromosome"
+  }
 
   #get maf data for a specific sample.
-  maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
+  if(missing(maf_data) && is.null(maf_path)){
+    maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
```
rdmorin, line 2775: What happens here if maf_path is provided? It looks like the code isn't handling that case.

R/viz.R
```diff
@@ -2819,8 +2849,23 @@ fancy_snv_chrdistplot = function(this_sample,
...
+  }else if (!is.null(maf_path)){
+    maf = fread_maf(maf_path)
+    maf = as.data.frame(maf)
+    colnames(maf)[variant_type_col] = "Variant_Type"
+    colnames(maf)[chromosome_col] = "Chromosome"
+  }
+
   #get maf data for a specific sample.
-  maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
+  if(missing(maf_data) && is.null(maf_path)){
```
rdmorin, line 2866: same comment as above

R/viz.R
```diff
@@ -2902,8 +2955,23 @@ fancy_svbar = function(this_sample,
...
+
+  }else if (!is.null(maf_path)){
+    maf = fread_maf(maf_path)
+    maf = as.data.frame(maf)
+    colnames(maf)[variant_type_col] = "Variant_Type"
+    colnames(maf)[chromosome_col] = "Chromosome"
+  }
+
   #get maf data for a specific sample.
-  maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
```
rdmorin, line 2906: same comment as above

R/viz.R
```diff
@@ -2952,14 +3026,39 @@ fancy_svbar = function(this_sample,
...
+    seq = as.data.frame(seq)
+    colnames(seq)[chrom_col] = "chrom"
+    colnames(seq)[start_col] = "start"
+    colnames(seq)[end_col] = "end"
+    colnames(seq)[cn_col] = "CN"
+  }
+
   #get maf data for a specific sample.
-  seq = get_sample_cn_segments(this_sample = this_sample, multiple_samples = FALSE, streamlined = FALSE, from_flatfile = TRUE)
+  if(missing(seq_data) && is.null(seq_path)){
```
rdmorin, line 3059: Same comment as above

R/viz.R
```diff
@@ -3045,15 +3150,40 @@ fancy_cnbar = function(this_sample,
...
+    maf = fread_maf(maf_path)
+    maf = as.data.frame(maf)
+    colnames(maf)[variant_type_col] = "Variant_Type"
+    colnames(maf)[chromosome_col] = "Chromosome"
+    colnames(maf)[start_col] = "Start_Position"
+    colnames(maf)[end_col] = "End_Position"
+  }
+
   #get maf data for a specific sample.
-  maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
```
rdmorin, line 3056: same comment

rdmorin (changes requested):
A few comments and questions

R/database.R
mattssca: Indeed, good catch!

R/viz.R
mattssca: I think it does, after some testing everything seems to be handled the right way:

```
#get maf as df in R.
maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = FALSE, from_flatfile = TRUE, use_augmented_maf = TRUE)$maf

#export maf (to read back into r with maf_path).
write.table(maf, "../../maf.maf", quote = FALSE, sep = "\t", row.names = FALSE, col.names = TRUE)

#using maf_path to build plot.
sv_plot_maf_path = fancy_sv_chrdistplot(this_sample = "", plot_subtitle = "maf_path to build plot", maf_path = "../../maf.maf")

#using maf_data to build plot.
sv_plot_maf_data = fancy_sv_chrdistplot(this_sample = "", plot_subtitle = "maf_data to build plot", maf_data = maf)

#using assign_cn_to_ssm to retrieve plotting data.
sv_plot_asssign_cn_to_ssm = fancy_sv_chrdistplot(this_sample = "HTMCP-01-06-00422-01A-01D", plot_title = "", plot_subtitle = "assign_cn_to_ssm to retrieve plotting data")
```
Screenshot 2022-05-10 at 12 26 40

R/viz.R
mattssca: same comment as above for me as well.

R/viz.R
mattssca: same comment as above for me as well.

R/viz.R
mattssca: same comment as above for me as well.

R/viz.R
mattssca: same comment as above for me as well.

mattssca (commented):
Thanks for your comments Ryan!

R/viz.R
```diff
@@ -2354,18 +2356,25 @@ prettyForestPlot = function(maf,
...
     mutmat = maf %>%
       dplyr::select(Hugo_Symbol, Tumor_Sample_Barcode) %>%
-      left_join(dplyr::select(metadata, Tumor_Sample_Barcode, comparison), by = "Tumor_Sample_Barcode") %>%
+      full_join(dplyr::select(metadata, Tumor_Sample_Barcode, comparison), by = "Tumor_Sample_Barcode") %>%
       distinct() %>%
       dplyr::mutate(is_mutated = 1) %>%
       pivot_wider(names_from = Hugo_Symbol, values_from = is_mutated, values_fill = 0) %>%
       dplyr::mutate(across(where(is.numeric), ~replace_na(., 0)))
+
+    if(rm_na_samples){
```
rdmorin, line 2372: I don't think I understand how this works. What's the format of the table you're working with here? I can't follow it from the code.

R/viz.R
rdmorin: Grat

rdmorin (commented):
Can you clarify how the code you added works, maybe with some in-line comments?

R/viz.R
mattssca: So, instead of a left join, mutmaf is now obtained with a full join (NA column retained). That temporary column is then renamed, and all rows that have 0 in that column are filtered out and the column is dropped (if the new parameter is set to true).  That being said, I am not 100% sure I understand what the issue is, so it might very well not be solving the issue that was raised.

rdmorin: To test this you would need to supply a metadata table that has more samples in it than your MAF does (in at least one of the two groups you want to compare). The two ways of running it would be:
default (not implemented in original version: use total samples in the metadata for that group as your denominator rather than the subset that are in the MAF
optional (actual current behaviour): use the subset in the MAF as the denominator and ignore the fact that there are more cases in your group in the metadata

mattssca: Ok, thanks for explaining this a bit better. I will test it out and make some changes as needed.

rdmorin: Since she wrote this function, @lkhilton can probably offer you a better review here than I if you get stuck.

R/viz.R
```diff
@@ -2313,6 +2313,7 @@ prettyForestPlot = function(maf,
                             metadata,
                             genes,
                             comparison_column,
+                            rm_na_samples = TRUE,
```
lkhilton, line 2316: Please make this FALSE by default

R/viz.R
```diff
@@ -2354,18 +2356,23 @@ prettyForestPlot = function(maf,
...
+      full_join(dplyr::select(metadata, Tumor_Sample_Barcode, comparison), by = "Tumor_Sample_Barcode") %>%
       distinct() %>%
       dplyr::mutate(is_mutated = 1) %>%
       pivot_wider(names_from = Hugo_Symbol, values_from = is_mutated, values_fill = 0) %>%
       dplyr::mutate(across(where(is.numeric), ~replace_na(., 0)))
+    if(rm_na_samples){ #remove 0 mutation samples, i.e samples that show 0 for all selected genes.
+      names(mutmat)[length(names(mutmat))]<-"na_samples" #rename the last column, "NA" (since dplyr do not treat a variable named NA the same way as other variables?)
+      mutmat = mutmat %>%
+        dplyr::filter(na_samples == 0) %>% #filter out all samples that show no mutations in the selected genes (i.e na_samples = 1).
+        dplyr::select(-na_samples) #drop temporary column (na_samples) to return mutmat to expected format.
```
lkhilton, line 2375: If you put this step (`select(-na_samples)`) outside this conditional subsetting, you can be sure to drop the `na_samples` column regardless of the value of `rm_na_samples`

mattssca: Fixed in [this](https://github.com/morinlab/GAMBLR/pull/88/commits/ba5f3a8078f305d9cc1931d199bca22a0aa92c9b) commit

rdmorin: Shall I approve this on my end?

mattssca: Yes, please.

R/viz.R
```diff
@@ -2354,18 +2356,23 @@ prettyForestPlot = function(maf,
...
+      full_join(dplyr::select(metadata, Tumor_Sample_Barcode, comparison), by = "Tumor_Sample_Barcode") %>%
       distinct() %>%
       dplyr::mutate(is_mutated = 1) %>%
       pivot_wider(names_from = Hugo_Symbol, values_from = is_mutated, values_fill = 0) %>%
       dplyr::mutate(across(where(is.numeric), ~replace_na(., 0)))
+    if(rm_na_samples && "NA" %in% colnames(mutmat)){
+      mutmat = mutmat %>%
+        dplyr::filter(`NA` == 0) #filter out all samples that show no mutations in the selected genes (i.e na_samples = 1).
+    }
+    mutmat = dplyr::select(mutmat, -`NA`)
```
lkhilton, line 2375: Sorry for missing this when we were chatting about this yesterday, but I think this still needs to be conditional because if the column `NA` isn't in the data frame this will error out. 
```
if("NA" %in% colnames(mutmat)){
    mutmat = dplyr::select(mutmat, -`NA`)
}
```

lkhilton (changes requested):

lkhilton (approved):

PR #91, opened 2022-05-18
Cmattsson dev

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


mattssca:
More to come this week, I'll ask for reviewers once the PR is ready to be merged. Thanks,

R/viz.R
```diff
@@ -2781,15 +2794,36 @@ fancy_sv_chrdistplot = function(this_sample,
...
+      maf = assign_cn_to_ssm(this_sample = this_sample, coding_only = coding_only, from_flatfile = from_flatfile, use_augmented_maf = use_augmented_maf)$maf
+    }else{
+      maf = get_combined_sv(sample_ids = this_sample, projection = projection, min_vaf = min_vaf) %>%
+        dplyr::select(CHROM_A, START_A, END_A, manta_name)
+
+      #get manta results in required format
+      maf = data.frame(maf$CHROM_A, maf$START_A, maf$END_A, do.call(rbind, strsplit(maf$manta_name, split = ":", fixed = TRUE)))
+
+      #rename variables
+      names(maf)[1] = "Chromosome"
```
Kdreval, line 2807: You can simplify here to
names(maf)[1:4] = c("Chromosome", "Start_Position", "End_Position","Variant_Type")

R/viz.R
```diff
@@ -2781,15 +2794,36 @@ fancy_sv_chrdistplot = function(this_sample,
...
+      names(maf)[2] = "Start_Position"
+      names(maf)[3] = "End_Position"
+      names(maf)[4] = "Variant_Type"
+
+      #filter out translocations and set order of variables
+      maf = dplyr::filter(maf, Variant_Type %in% c("MantaDEL", "MantaDUP")) %>%
```
Kdreval, line 2813: This will probably drop SVs coming not from manta, but other sources (like GRIDSS)?

R/database.R
```diff
@@ -1346,22 +1396,24 @@ get_ashm_count_matrix = function(regions_bed,
...
-                                from_indexed_flatfile = from_indexed_flatfile,
-                                allow_clustered = allow_clustered)
+                                from_indexed_flatfile = from_indexed_flatfile)
 
   ashm_counted = ashm_maf %>%
-    group_by(sample_id, region_name) %>%
+    group_by(Tumor_Sample_Barcode, Region_Name) %>%
     tally()
+
+  colnames(ashm_counted)[1] = "sample_id"
```
Kdreval, line 1414: This does not seem to do anything because the next line re-assigns the name of this column

R/database.R
```diff
@@ -1346,22 +1396,24 @@ get_ashm_count_matrix = function(regions_bed,
...
+  colnames(ashm_counted)[2] = "region_name"
```
Kdreval, line 1415: The `get_ssm_by_regions` below specifies this to be `colnames(r_names)[1] <- "Region_Name"`. So basically in the internal function we are assigning one name, and here re-naming it. Will it make sense to assign names in one place and then use it consistently? What is the default name, maybe we do not need to re-name it at all? I'll leave a comment at the line that I mean

R/database.R
```diff
@@ -1430,40 +1488,59 @@ get_ssm_by_regions = function(regions_list,
...
   }
+
+  #get region names
   tibbled_data = tibble(region_mafs, region_name = rn)
   unnested_df = tibbled_data %>%
     unnest_longer(region_mafs)
 
+  #subset on region anmes and rename column before cbind with full maf
+  r_names = as.data.frame(unnested_df$region_name)
+  colnames(r_names)[1] <- "Region_Name"
```
Kdreval, line 1518: Here, this line is related to the comment above. What is the default name of first column here?

mattssca:
Also added updates to prettyOncoplot and annotate_hotspots provided by Kostia. package documentation has been regenerated.

R/database.R
```diff
@@ -1520,29 +1599,28 @@ get_ssm_by_region = function(chromosome,
         maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$augmented
       }else{
         maf_partial_path = config::get("results_flatfiles")$ssm$template$merged$deblacklisted
-        }
+      }
     }else if (mode == "strelka2"){
       maf_partial_path = config::get("results_flatfiles")$ssm$all$strelka2
```
rdmorin, line 1604: In the section above we need to get rid of the hard-coded path. Can you put this path in a new section in the config file?

tabix_bin = "/home/rmorin/miniconda3/bin/tabix"

e.g. 
default:
    dependencies:
        tabix: "/home/rmorin/miniconda3/bin/tabix"

mattssca: Will do!

R/database.R
```diff
@@ -65,7 +71,7 @@ get_ssm_by_patients = function(these_patient_ids,
     ungroup()
 
   these_sample_ids = pull(these_samples_metadata, sample_id)
-  return(get_ssm_by_samples(these_sample_ids, these_samples_metadata, tool_name, projection, seq_type, flavour, min_read_support, subset_from_merge))
+  return(get_ssm_by_samples(these_sample_ids, these_samples_metadata, tool_name, projection, seq_type, flavour, min_read_support, subset_from_merge, basic_columns, maf_cols, return_cols))
```
rdmorin, line 74: It's probably safer to pass all these as named arguments in case the order or number of arguments in get_ssm_by_samples ever changes

R/database.R
```diff
@@ -95,6 +104,9 @@ get_ssm_by_samples = function(these_sample_ids,
                               flavour = "clustered",
                               these_genes,
                               min_read_support = 3,
+                              basic_columns = TRUE,
+                              maf_cols = NULL,
+                              return_cols = FALSE,
```
rdmorin, line 109: I understand the convenience option here but I wonder if this is better placed in a separate function. It could even be a global variable "maf_cols" if it was consistent or could be set as a global variable the first time this function or some other MAF-loading function is called.

rdmorin: I'm envisioning a way for the user could set a global variable that is a list of their preferred columns by subsetting from the full global variable. If that was ensured to exist then the function could make use of it directly from their environment without even asking for it.

R/database.R
```diff
@@ -137,6 +155,12 @@ get_ssm_by_samples = function(these_sample_ids,
       maf_df_merge = fread_maf(full_maf_path) %>%
         dplyr::filter(Tumor_Sample_Barcode %in% these_sample_ids) %>%
         dplyr::filter(t_alt_count >= min_read_support)
+      #subset maf to only include first 43 columns (default)
+      if(basic_columns){maf_df_merge = dplyr::select(maf_df_merge, c(1:45))}
+      #subset maf to a specific set of columns (defined in maf_cols)
+      if(!is.null(maf_cols) && !basic_columns){maf_df_merge = dplyr::select(maf_df_merge, all_of(maf_cols))}
+      #print all available columns
```
rdmorin, line 162: If this was implemented to save time the way it is set up probably doesn't help much because it still loads the full MAF. We need a way to "peek" at (i.e. head) the file for whatever approach we use to get the column names

R/database.R
rdmorin: Which line are you saying doesn't do anything? 1412-1414?

R/database.R
```diff
@@ -1432,40 +1488,59 @@ get_ssm_by_regions = function(regions_list,
...
+  colnames(r_names)[1] <- "region_name"
+
+  #unnest maf
+  unnested_maf = bind_rows(region_mafs)
+
+  #join region names df with unnested maf
+  maf = cbind(r_names, unnested_maf)
+
   if(streamlined){
-    unlisted_df = mutate(unnested_df, start = region_mafs$Start_Position, sample_id = region_mafs$Tumor_Sample_Barcode) %>%
```
rdmorin, line 1459: I'm very confused by this change and how it maintains the original functionality

rdmorin (changes requested):

R/database.R
mattssca: There might be a more elegant solution to this. My update achieves the same output in the following steps:

1. Subsets region names from the unnested df
2. Renames the column (to region_name, to match downstream required format)
3. Binding maf df (unnests)
4. Joins the df with region names with unnested maf. 
 
The integrity of the final returned maf is retained (at least to the testing I've done). Does this clarify the confusion?

R/database.R
mattssca: The renaming of the column is necessary since the unnesting step gives it a different name (dfname$columnname). The renaming was reverted to "region_name" to match what's expected downstream and the redundant renaming step was removed (in get_ashm_count_matrix).

mattssca (commented):
Thanks for your review comments. Please see my questions related to how to proceed with maf-column-subset and let me know if there are any outstanding questions. 

Also, If someone could have a look at the newly added collate function (colalte_qc_results) that would also be good in preparation to get this PR merged into master. I am assuming merge conflicts would have to be tackled as the last step and potentially after the other PR (GAMBLR remote) has been merged into master.

mattssca:
FYI, After merging this branch with rmorin-dev, the following warnings were returned when generating package documentation: 
```
Warning: [/home/cmattsson/GAMBLR/cmattsson-dev/GAMBLR/R/checks.R:15] @examples requires a value
Warning: [/home/cmattsson/GAMBLR/cmattsson-dev/GAMBLR/R/utilities.R:13] @examples requires a value
Warning: [/home/cmattsson/GAMBLR/cmattsson-dev/GAMBLR/R/utilities.R:1067] @param requires name and description
Warning: [/home/cmattsson/GAMBLR/cmattsson-dev/GAMBLR/R/utilities.R:1073] @examples requires a value
```

R/database.R
```diff
@@ -13,38 +13,49 @@ cnames = c("CHROM_A", "START_A", "END_A", "CHROM_B", "START_B", "END_B", "NAME",
 #' excluded_samp = get_excluded_samples()
 #'
```
rdmorin, line 14: Not sure who added this but it needs documentation

R/database.R
rdmorin: I think it's fine for now as long as it works as advertised :)

R/database.R
rdmorin: Exactly

rdmorin: The problem though is we need the NUMERIC columns to give to cut because it can't select the columns based on their names. We can leave this as a future feature.

R/viz.R
rdmorin: The non-Manta ones may not have annotations in that column. Leave it for now. This can be a TO DO

rdmorin (approved):

PR #92, opened 2022-05-19
init prettyCoOncoplot

Kdreval:
Adding the functionality to show 2 prettyOncoplots side-by-side. The output is ggplot-compatible and can be arranged together with other plots in a multi-panel figure.


R/viz.R
```diff
@@ -1624,6 +1624,112 @@ prettyOncoplot = function(maftools_obj,
...
+#' @return A ggplot object with 2 oncoplots side-by-side.
+#' @export
+#' @import ComplexHeatmap ggpubr
+#'
+#' @examples
+#' ssm1=get_coding_ssm(limit_cohort = "BL_Adult")
+#' ssm1=maftools::read.maf(ssm1)
+#' ssm2=get_coding_ssm(limit_cohort = "BL_Pediatric")
+#' ssm2=maftools::read.maf(ssm2)
+#' meta1=get_gambl_metadata() %>% filter(cohort=="BL_Adult")
```
mattssca, line 1655: Not sure how important this is, but when I tried to run the example I got an error saying "Error in filter(., cohort == "BL_Pediatric") : object 'cohort' not found ". This is since the package related to the intended filter function is not specified. Adding dplyr:: in front of the filter statement solved it and the example runs. Just a minor comment. The rest of the example works great and I now have a side-by-side (very) Pretty Oncoplot :)

R/viz.R
```diff
@@ -1624,6 +1624,112 @@ prettyOncoplot = function(maftools_obj,
...
+
+    # Check for required arguments
+    required = c("ssm1", "ssm2", "meta1", "meta2")
+    defined = names(as.list(match.call())[-1])
+    if (any(!required %in% defined)) {
+      stop("Please provide mutation data and metadata for 2 pretty Oncoplots.")
+    }
+
+    # Arguments to pass into prettyOncoplot
+    oncoplot_args = list(...)
```
mattssca, line 1685: This is great!

NAMESPACE
```diff
@@ -104,6 +105,7 @@ import(dbplyr)
 import(dplyr)
 import(g3viz)
 import(ggplot2)
+import(ggpubr)
```
mattssca, line 108: In regards to ggpubr, do you want to update the DESCRIPTION as well?

R/utilities.R
```diff
@@ -1725,11 +1725,11 @@ get_gambl_colours = function(classification = "all",
                             "NEG" = "#E5A4CB")
 
   all_colours[["BL"]] = c("M53-BL" = "#A6CEE3",
-                          "DLBCL-1" = "#721F0F",
+                          "DLBCL-A" = "#721F0F",
```
mattssca, line 1728: Just curious, what's the rationale for moving away from numbers to letters in this case?

mattssca (commented):
Awesome addition to GAMBLR.! Thanks, Kostia. Just a few minor comments/questions.

R/database.R
```diff
@@ -1093,8 +1093,10 @@ get_cn_states = function(regions_list,
   cn_matrix = pivot_wider(all_cn, id_cols = "sample_id", names_from = "region_name", values_from = "CN") %>%
     column_to_rownames("sample_id")
 
+  names(cn_matrix) = region_names
```
rdmorin, line 1096: Why was this changed? Is it a bug fix? It seems like it will affect the output.

R/viz.R
Kdreval: Yes, this happens for me too when I run `devtools::load_all()`. You are right, the example should say `dplyr::filter` - I will correct this and re-generate documentation

R/viz.R
```diff
@@ -1624,6 +1624,112 @@ prettyOncoplot = function(maftools_obj,
...
+#'     include_noncoding = NULL,
+#'     minMutationPercent = 0,
+#'     genes=c("MYC", "TET2", "TP53", "DDX3X", "ID3"),
+#'     metadataColumns=c("pathology", "EBV_status_inf", "pairing_status", "cohort"),
+#'     splitColumnName="EBV_status_inf",
+#'     metadataBarHeight=10,
+#'     fontSizeGene=12,
+#'     metadataBarFontsize=10,
+#'     label1="Adult",
+#'     label2="Pediatric")
```
rdmorin, line 1667: I have a suggestion for how to make this function easier to use. What about implementing it the same way prettyForestPlot is set up. The user doesn't have to worry about subsetting. They just provide one mutation_df and one metadata table and specify the column and two values to use to subset the data to the two groups. Would this have any drawbacks?

rdmorin (changes requested):
Some thoughts to consider before we lock this in.

R/utilities.R
Kdreval: Ah, this is because we will rename these subgroups as part of our response to reviewer's comments fr our manuscript. The concern is that the clusters C1 to C5 already exist for DLBCL, and this may create confusion

R/database.R
Kdreval: Yes, I had this not working as it was before. I think it might be related to R v3.6 to R v4 shift. Basically it was returning an unnamed vector of numbers (CN states) instead of a named column in a df

Kdreval:
@mattssca , @rdmorin : Thank you s much for the reviews! The comments were addressed and pushed to git. This is now ready for re-review

rdmorin (commented):
Nice addition!

rdmorin (approved):

Kdreval:
Thank you both for review and approval!

PR #96, opened 2022-05-26
update collate functions to work for genome and capture data 

rdmorin:
…e each result separately# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1522,8 +1522,9 @@ collate_sbs_results = function(sample_table,
   if(missing(file_path)){
     base = config::get("project_base")
 
-    file_path = paste0(base,"results/icgc_dart/sigprofiler-1.0/02-extract/genome--hg38/BL_HGBL_DLBCL_FL_COMFL_CLL_MCL_B-ALL_PBL_DLBCL-BL-like_UNSPECIFIED_SCBC_MM_all/SBS96/Suggested_Solution/COSMIC_SBS96_Decomposed_Solution/Activities/COSMIC_SBS96_Activities_refit.txt")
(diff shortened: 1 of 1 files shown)


rdmorin:
One important note here is that I had to update some functions that were calling get_gambl_metadata without passing along a seq_type_filter parameter. Every function that does this needs to be updated so it will work with capture data. There was also a call to get_ssm_by_region I had to fix (same reason). I suspect there are others still in need of this patch.

R/utilities.R
```diff
@@ -1571,22 +1592,27 @@ collate_sbs_results = function(sample_table,
 #' @examples
 #' sample_table = collate_nfkbiz_results(sample_table=sample_table)
 #'
-collate_nfkbiz_results = function(sample_table){
-
+collate_nfkbiz_results = function(sample_table,seq_type_filter="genome"){
+  #TO DO: Update to work with hg38 projection
```
mattssca, line 1596: I created an issue of this, to allow for better tracking.

mattssca (approved):
Thanks Ryan. For our convenience, I've extracted all GAMBLR functions that, to some extent call get_gambl_metadata, so that we can more efficiently have a look at these individual functions if they need to be updated in a similar way. The functions are: 
> get_sample_cn_segments
> get_ashm_count_matrix
> get_coding_ssm
> get_gene_cn_and_expression
> get_gene_expression
> get_ssm_by_patients
> get_ssm_by_samples
> get_ssm_by_sample
> get_coding_ssm_status
> calc_mutation_frequency_sliding_windows
> sv_to_custom_track 
> maf_to_custom_track
> collate_results
> get_bams
> cnvKompare
> plot_sample_circos
> ashm_multi_rainbow_plot
> populate_each_tool_result

I've also moved one newly added "TODO" to GAMBLR issues.

PR #99, opened 2022-06-02
Remote GAMBLR (outside GSC) is now a thing. Testers wanted!

rdmorin:
This was tested on my laptop using the test_remote.R script bundled with GAMBLR. The snakefile to allow the user to sync the required files is also provided. I haven't made a yml file for that environment yet. 

# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


rdmorin:
This now has a patch for cBIoportal (portal.R) that fixes some issues caused by recent updates.

R/checks.R
```diff
@@ -0,0 +1,86 @@
...
+  
+  if(to_check == "all"){
+    to_check = c("uniqueness","colours","completeness")
+  }
+  if("completeness" %in% to_check){
+    if(any(!required_cols %in% colnames(metadata_df))){
+      missing = required_cols[!required_cols %in% colnames(metadata_df)]
+      message("FAIL! MISSING metadata columns:")
+      print(paste(missing,sep=" "))
+      return()
```
Kdreval, line 26: What about adding here `else` part that will just say something alongside "Success! All required columns are present in the metadata". This way we are letting user know that the completeness check was performed and all is well

R/viz.R
```diff
@@ -1,4 +1,12 @@
 
+# global set of aliases for finding specific sets of colours
+colour_aliases = list("COO_consensus" = "coo", "COO" = "coo", "DHITsig_consensus" = "coo",
+                      "pathology" = "pathology", "analysis_cohort" = "pathology", "group" = "pathology",
+                      "FL_group" = "FL", "lymphgen" = "lymphgen", "lymphgen_with_cnv" = "lymphgen",
+                      "bcl2_ba" = "pos_neg", "BCL2_status" = "pos_neg", "myc_ba" = "pos_neg",
+                      "bcl6_ba" = "pos_neg", "manta_BCL2_sv" = "pos_neg", "manual_BCL2_sv" = "pos_neg", "manta_MYC_sv" = "pos_neg")
```
Kdreval, line 7: One other thing. I can think of adding here is `"EBV_status_inf"="EBV_status"`

data-raw/DATASET.R
```diff
@@ -91,3 +91,52 @@ usethis::use_data(chromosome_arms_grch37, overwrite = TRUE)
...
+gene_detail[gene_detail$ensembl_gene_id =="ENSG00000102096","entrezgene_id"] = 11040
+
+gene_detail[gene_detail$ensembl_gene_id =="ENSG00000172578","entrezgene_id"] = 89857
+
+gene_detail[gene_detail$ensembl_gene_id =="ENSG00000205542","entrezgene_id"] = 7114
+
+lymphoma_genes = left_join(lymphoma_genes,gene_detail) 
+lymphoma_genes$LymphGen=FALSE
+lymphoma_genes[lymphoma_genes$entrezgene_id %in% lymphgen_anno$NCBI_Gene_ID,"LymphGen"] = TRUE
+
```
Kdreval, line 129: It seems that here line
```usethis::use_data(lymphoma_genes, overwrite = TRUE)```
might be missing?

File: get_gambl_results.smk [hunk 4]
```diff
@@ -0,0 +1,80 @@
...
+    run:
+        shell("cp {input[0]} {output[0]}")
+
+rule get_sv:
+    input:
+        SFTP.remote(hostname + project_base + sv_combined)
+    output:
+        sv_combined
+    run:
+        shell("cp {input[0]} {output[0]}")
```
Kdreval, line 80: New line is missing here, but would it make sense to also add the expression data syncing here?

Kdreval (approved):

mattssca (approved):
This works as intended on Rstudio (remote, gphost) for me. Please see the comments in the review for a few questions and thoughts. 

I will give this a go on my local machine as well (using snakefile to get all required files). 

As for cBioPortal study setup, importing the genome study works fine (currently added to our instance), but for some strange reason (Docker being Docker...), it does not work for the capture study. I will troubleshoot this and it's very likely that this error is related to Docker and not the code in this PR. So should be fine to approve this PR.

edit: The capture study was successfully imported into our cBioPortal instance. This was achieved by increasing the memory allocated to Docker.

test_remote.R
```diff
@@ -1,16 +1,145 @@
...
+all_ssm = bind_rows(coding_ssm,cap_ssm)
+
+
+blacklisted = GAMBLR:::annotate_ssm_blacklist(cap_ssm_hg38,seq_type = "capture",genome_build = "hg38")
+
+#test SV functions
+svar_all = get_combined_sv(min_vaf = 0.1,projection = "grch37")
+
+dim(svar_all)
+# [1] 450641     19
```
lkhilton, line 57: This returned a very different value for me: 
```
[1] 150162     19
```

test_remote.R
```diff
@@ -1,16 +1,145 @@
...
+
+#test annotation
+svar_anno = annotate_sv(svar_all)
+
+
+# test CN functionality
+
+all_seg = get_sample_cn_segments(sample_list = all_meta$sample_id,multiple_samples = T)
+#dim(all_seg)
+#[1] 375247      7
```
lkhilton, line 67: This returned the following: 
```
[1] 376588      7
```

test_remote.R
```diff
@@ -1,16 +1,145 @@
...
+#14-24534_tumorA 14-24534_tumorB 
+#11097           12904 
+
+
+#copy number functionality
+
+cn_seg = get_sample_cn_segments("14-24534_tumorB",from_flatfile = T)
+
+#should automagically get the single seg file and MAF for the sample
+cn_ssm = assign_cn_to_ssm("14-24534_tumorB",seg_file_source = "battenberg",ssh_session = session)
```
lkhilton, line 96: This gives me the following error: 
```
trying to find output from: battenberg                                                                      
looking for flatfile: /projects/nhl_meta_analysis_scratch/gambl/results_local/
~/gambl/results_local/
Error in file(con, "rb") : cannot open the connection
In addition: Warning messages:
1: In file(con, "rb") :
  'raw = FALSE' but '/Users/laurahilton/gambl/results_local' is not a regular file
2: In file(con, "rb") :
  cannot open file '/Users/laurahilton/gambl/results_local': it is a directory
```

test_remote.R
```diff
@@ -1,16 +1,145 @@
...
+
+
+pursteenah = estimate_purity(sample_id = "14-24534_tumorB",seg_file_source = "battenberg",ssh_session=session)
+
+pursteenah$sample_purity_estimation
+
+# advanced functions
+
+
+coding_ssm_status = get_coding_ssm_status(gene_symbols = c("EZH2","CREBBP","KMT2D"),these_samples_metadata = cap_meta,maf_data = cap_ssm)
```
lkhilton, line 106: This gives me a similar error message so it's not related to the ssh connection? 
```
Joining, by = "sample_id"
Error in file(con, "rb") : cannot open the connection
In addition: Warning messages:
1: In file(con, "rb") :
  'raw = FALSE' but '/Users/laurahilton/gambl' is not a regular file
2: In file(con, "rb") :
  cannot open file '/Users/laurahilton/gambl': it is a directory
```

test_remote.R
```diff
@@ -1,16 +1,145 @@
...
+
+
+
+# test if we can make a cBioportal instance
```
lkhilton, line 110: I deleted Docker from my laptop today so I'm unhelpful here

lkhilton (commented):
I ran into some errors when running test_remote.R outlined in my comments. I also committed a couple of changes I caught in the process of testing.

test_remote.R
rdmorin: No worries. This part has been tested by Adam.

test_remote.R
rdmorin: I get 450641 on the gphost too. Must be a truncated file on your end.

test_remote.R
rdmorin: Yeah I get that number now too.

R/utilities.R
```diff
@@ -9,11 +9,15 @@ names(rainfall_conv) = c('A>G', 'T>C', 'C>T', 'G>A', 'A>T', 'T>A', 'A>C', 'T>G',
 #'
 #' @return
 #' @export
-#' @import ssh
 #'
 #' @examples
 get_ssh_session = function(host="gphost01.bcgsc.ca"){
-  session = ssh_connect(host=host)
+  if (!requireNamespace("ssh", quietly = TRUE)) {
```
lkhilton, line 15: I added `library(ssh)` to the top of the `test_remote.R` script because without it in NAMESPACE this throws an error. Is it possible to also add a step to load the library if it's already installed?

rdmorin: Good idea. I suppose we could have it load automatically but I think it's frowned upon to put library() or require() into functions like that.

rdmorin: Since this function only needs to be run once per "session" I figured this warning would be fine

rdmorin: Oh. It's not printing this as a message?
"The ssh package must be installed to use this functionality"

PR #103, opened 2022-06-15
new feature: right annotations in prettyOncoplot

Kdreval:
Small update to support right side annotation tracks based on `prettyForestpot` outputs to label genes significantly enriched in one or the other pathology.
Example:
image


mattssca (approved):
Nice new feature! Thanks, Kostia. Would we want to add an @example of this new feature here? Alternatively, this could be highlighted in the upcoming vignette (I'm fine with the latter).

PR #104, opened 2022-06-22
Add additional blacklist to annotate_ssm_blacklist()

hayashaalan:
I added Chris' curated blacklist to the annotate_ssm_blacklist function. I've also created a GAMBL PR to get that blacklist onto master.

# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


mattssca:
@hayashaalan Do want to resolve the merge conflict? I can then review this PR and merge it into master.

Thanks,

hayashaalan:
I've resolved the merge conflict. Thanks Adam!

config.yml
```diff
@@ -27,6 +27,7 @@ default:
         oncodriveclustl:
                 clusters: "versioned_results/oncodriveclustl-0.0_{abase}_clusters_results.tsv"
                 elements: "versioned_results/oncodriveclustl-0.0_{abase}_elements_results.tsv"
+        curated_blacklist: "versioned_results/curated_blacklist/candidateblacklist.schmitz.{projection}.tsv"
```
Kdreval, line 30: @hayashaalan is this file already in GAMBL? Just want to make sure it is on master so everyone has access to it and then this can be merged. Thanks!

hayashaalan: I have a GAMBL PR that contains the necessary files!

R/annotation.R
```diff
@@ -80,7 +81,37 @@ annotate_ssm_blacklist = function(mutations_df,
   mutations_df = left_join(mutations_df,combined_blacklist,by = c("Chromosome", "Start_Position")) %>%
     mutate(blacklist_count = replace_na(blacklist_count, 0))
 
+  if (use_curated_blacklist){
+    repo_base = config::get("repo_base")
+    full_path = paste0(repo_base, config::get("results_versioned")$curated_blacklist)
+    additional_blacklist = glue(repo_path) %>% read_tsv(c("chrpos", "blacklist_count2"), show_col_types = FALSE)
```
rdmorin, line 87: I think it's better to have a header in your file so that we don't need to hardcode this in GAMBLR

rdmorin (changes requested):
See my comments regarding header and hardcoding

config.yml
```diff
@@ -27,6 +27,8 @@ default:
         oncodriveclustl:
                 clusters: "versioned_results/oncodriveclustl-0.0_{abase}_clusters_results.tsv"
                 elements: "versioned_results/oncodriveclustl-0.0_{abase}_elements_results.tsv"
+    etc:
+      curated_blacklist: "etc/curated_blacklist/candidateblacklist.schmitz.{projection}.tsv"
     resources:
```
rdmorin, line 32: I suggest this be moved to the key "resources" e.g.

```
resources:
    curated_blacklist: "etc/curated_blacklist/candidateblacklist.schmitz.{projection}.tsv"
```

R/annotation.R
```diff
@@ -80,7 +81,36 @@ annotate_ssm_blacklist = function(mutations_df,
   mutations_df = left_join(mutations_df,combined_blacklist,by = c("Chromosome", "Start_Position")) %>%
     mutate(blacklist_count = replace_na(blacklist_count, 0))
 
+  if (use_curated_blacklist){
+    repo_base = config::get("repo_base")
+    full_path = paste0(repo_base, config::get("etc")$curated_blacklist)
```
rdmorin, line 86: Change this to match the change I suggested below
```
config::get("resources")$curated blacklist
```

R/annotation.R
```diff
@@ -80,7 +81,36 @@ annotate_ssm_blacklist = function(mutations_df,
...
+    additional_blacklist = glue(full_path) %>% read_tsv()
+    additional_blacklist = additional_blacklist %>%
+      separate(chrpos, into = c("Chromosome", "Start_Position"), sep = ":")
+
+    additional_blacklist = mutate(additional_blacklist, Start_Position = as.integer(Start_Position)) %>%
+      group_by(Start_Position, Chromosome) %>%
```
rdmorin, line 92: Why are you grouping and summarizing here? Isn't that already dealt with in your blacklist?

R/annotation.R
```diff
@@ -80,7 +81,36 @@ annotate_ssm_blacklist = function(mutations_df,
...
+      summarize(blacklist_schmitz = sum(blacklist_schmitz)) %>%
+      ungroup()
+
+    if(str_detect(mutations_df$Chromosome, "chr")[1]){
```
rdmorin, line 96: unnecessary if your black lists have the right prefixing for the projection. I'd leave this out and test that it works

R/annotation.R
```diff
@@ -80,7 +81,36 @@ annotate_ssm_blacklist = function(mutations_df,
...
+      additional_blacklist = mutate(additional_blacklist, Chromosome = paste0("chr", Chromosome))
+    }
+    mutations_df = left_join(mutations_df,additional_blacklist,by = c("Chromosome", "Start_Position")) %>%
+      mutate(blacklist_schmitz = replace_na(blacklist_schmitz, 0))
+    dropped = dplyr::filter(mutations_df, blacklist_schmitz > drop_threshold)
+    mutations_df = dplyr::filter(mutations_df, is.na(blacklist_schmitz) | blacklist_schmitz < drop_threshold)
+    if(verbose){
+      if(length(dropped) > 0 ){
```
rdmorin, line 104: Since it's a data frame this should be nrow(dropped) right?

rdmorin: You could also declare ndrop earlier and just check ndrop > whatever here

R/annotation.R
```diff
@@ -31,6 +31,7 @@ annotate_ssm_blacklist = function(mutations_df,
                                   blacklist_file_template,
                                   drop_threshold = 4,
                                   return_blacklist = FALSE,
+                                  use_curated_blacklist = FALSE,
```
rdmorin, line 34: Currently there's an issue with the logic here because I think the curated blacklist will be applied after the regular blacklist and the contents of `dropped` will be clobbered. It would probably be better if you set this up such that using the curated blacklist caused the earlier code involving the other blacklists to be skipped. e.g.

```
#run any code that is needed in all cases, e.g. setting projection from genome_build. 
if( ! use_curated_blacklist{
 do all the regular blacklist stuff and create/populate the two data frames to return them later
}else{
 do your new blacklist stuff and create/populate the the two data frames to return them outside this block
}
#code that runs in all cases:
if(invert){
    return(dropped)
  }
  return(mutations_df)
```

rdmorin (changes requested):

R/annotation.R
```diff
@@ -46,51 +47,69 @@ annotate_ssm_blacklist = function(mutations_df,
...
+    dropped = dplyr::filter(mutations_df, blacklist_count > drop_threshold)
+    if(verbose){
+      if(nrow(dropped) > 0 ){
+        ndrop = length(dropped$Tumor_Sample_Barcode)
+        message(paste(ndrop, "variants were dropped"))
+      } else {
+        message("0 variants were dropped")
+      }
+      
+    }
```
rdmorin, line 89: Fix indenting to match the block

rdmorin (changes requested):
Can you make sure the indenting of all braces is consistent? I can't tell where blocks begin and end right now and the diff is showing a missing closing brace at the end, which is worrisome

R/annotation.R
```diff
@@ -46,51 +47,66 @@ annotate_ssm_blacklist = function(mutations_df,
...
+    repo_base = config::get("repo_base")
+    full_path = paste0(repo_base, config::get("resources")$curated_blacklist)
+    additional_blacklist = glue(full_path) %>% read_tsv()
+    additional_blacklist = additional_blacklist %>%
+      separate(chrpos, into = c("Chromosome", "Start_Position"), sep = ":") %>% mutate(Start_Position = as.numeric(Start_Position))
+
+    mutations_df = left_join(mutations_df,additional_blacklist,by = c("Chromosome", "Start_Position")) %>%
+      mutate(blacklist_count = replace_na(blacklist_count, 0))
+    dropped = dplyr::filter(mutations_df, blacklist_count > drop_threshold)
+    mutations_df = dplyr::filter(mutations_df, is.na(blacklist_count) | blacklist_count < drop_threshold)
```
rdmorin, line 99: You can delete this line as it's redundant with line 111, which covers both cases now.

PR #110, opened 2022-07-04
New plotting functions and various improvements. 

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


mattssca:
More chances to come, will notify you when PR is ready for review. Thanks,

R/viz.R
```diff
@@ -2926,7 +2928,9 @@ fancy_v_chrcount = function(this_sample,
                             chr_select = paste0("chr", c(1:22)),
                             coding_only = FALSE,
                             from_flatfile = TRUE,
-                            use_augmented_maf = TRUE){
+                            use_augmented_maf = TRUE,
+                            add_qc_metric = FALSE,
+                            seq_type = "genome"){
```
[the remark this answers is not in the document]
mattssca: This parameter was added to satisfy the call of `collate_qc_results`. But I should also update the call of `assign_cn_to_ssm` to use the same `seq_type`. Thanks!

R/viz.R
```diff
@@ -4367,37 +4383,46 @@ fancy_sv_sizedens = function(this_sample,
...
+                                plot_sub = ""){
 
-  if(missing(these_samples) && !is.null(limit_cohort)){
-    these_samples = dplyr::filter(metadata, cohort == limit_cohort) %>%
-      dplyr::select(sample_id) %>%
-      arrange(sample_id) %>%
-      as.data.frame()
+  #get gambl metadata (if not supplied)
+  if(missing(metadata)){
+    this_meta = get_gambl_metadata()
```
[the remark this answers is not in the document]
mattssca: Indeed, thanks for your comment. All metadata functions used in these plotting functions will be updated to call `seq_type_filter`.

R/viz.R
```diff
@@ -4367,37 +4383,46 @@ fancy_sv_sizedens = function(this_sample,
...
+    this_meta = metadata
   }
 
-  if(missing(these_samples) && !is.null(limit_pathology)){
-    these_samples = dplyr::filter(metadata, pathology == limit_pathology) %>%
-      dplyr::select(sample_id) %>%
-      arrange(sample_id) %>%
-      as.data.frame()
+  #filter metadata on selected cohort (if filter_cohort is called)
+  if(missing(these_samples)){
```
rdmorin, line 4416: This will not work as expected if someone supplies BOTH filter_cohort AND filter_pathology, will it?

mattssca: No, it will not. The way I designed this function was so that the user can define one or the other. But I can see how it could be useful to further subset your data based on both conditions. I will address this.

R/viz.R
```diff
@@ -4522,18 +4548,19 @@ fancy_qc_plot = function(these_samples,
...
-      dplyr::select(sample_id) %>%
-      as.data.frame()
-  }
-
-  #filter metadata on pathology (if filter_pathology is called)
-  if(missing(these_samples) && !missing(filter_pathology)){
-    these_samples = dplyr::filter(this_meta, pathology == filter_pathology) %>%
-      dplyr::select(sample_id) %>%
-      as.data.frame()
+  if(missing(these_samples)){
```
rdmorin, line 4551: Same issue as above. Is the idea here that someone will only ever be expected to filter using one or the other but never combine them? If so, then you will need to check that they never provide more than one filter and throw an error if they do

mattssca: Yes, see my response to your previous comment. It's the same for all these functions. I think it's best to have the function filter on both if both parameters are used.

R/viz.R
```diff
@@ -4614,7 +4641,117 @@ fancy_qc_plot = function(these_samples,
...
+                              metadata,
+                              filter_cohort,
+                              filter_pathology,
+                              comparison_samples,
+                              seq_type = "genome",
+                              plot_subtitle = ""){
+
+  #get gambl metadata (if not supplied)
+  if(missing(metadata)){
+    this_meta = get_gambl_metadata()
```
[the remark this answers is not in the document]
mattssca: Noted, the full collection of these fancy QC plots needs to be updated in this regard.

mattssca:
Thanks for your comments Ryan. I completely understand the importance of the issues you raised and also that it is important to call get_gambl_metadata with the correct parameters. I am continuously making updates to these functions over the last couple of days and as the [comment ](https://github.com/morinlab/GAMBLR/pull/110#issuecomment-1174338942)on this PR states, It is not 100% for review yet. All good comments though, I will address these issues in an upcoming commit. Please do not worry.

Kdreval (approved):

rdmorin (approved):

Issue #112, opened 2022-07-13
`collate_results` has unexpected behaviour on join_with_full_metadata


lkhilton:
The documentation for this function implies that the `join_with_full_metadata` argument can be toggled to join with all metadata columns. However the function joins with all metadata columns from `get_gambl_metadata()` instead of from the user-supplied metadata table, so setting `join_with_full_metadata = TRUE` when running this function ignores the user-supplied metadata table. 

```
if (join_with_full_metadata) {
    # INSERT ANOTHER IF STATEMENT HERE TO USE THE USER_PROVIDED METADATA TABLE IF IT EXISTS
    full_meta = get_gambl_metadata(seq_type_filter = seq_type_filter)
    full_table = left_join(full_meta, sample_table)
    full_table = full_table %>% mutate(MYC_SV_any = case_when(ashm_MYC > 
      3 ~ "POS", manta_MYC_sv == "POS" ~ "POS", ICGC_MYC_sv == 
      "POS" ~ "POS", myc_ba == "POS" ~ "POS", TRUE ~ "NEG"))
    full_table = full_table %>% mutate(BCL2_SV_any = case_when(ashm_BCL2 > 
      3 ~ "POS", manta_BCL2_sv == "POS" ~ "POS", ICGC_BCL2_sv == 
      "POS" ~ "POS", bcl2_ba == "POS" ~ "POS", TRUE ~ 
      "NEG"))
    full_table = full_table %>% mutate(DoubleHitBCL2 = ifelse(BCL2_SV_any == 
      "POS" & MYC_SV_any == "POS", "Yes", "No"))
    return(full_table)
  }
```

PR #115, opened 2022-08-04
Fancy Plots + QoL Updates

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

- [x] My function uses a library that isn't already a dependency of GAMBLR and I made the package aware of this dependency using the function documentation `import` statment. 

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


mattssca:
This PR includes updates to a collection of fancy plots. The following revisions have been made.

1. fancy_cnbar - Remove lines connecting the dots.

2. fancy_ideogram - New parameter that allows the user to subset calls to regions specified in a bed-like format. For more info and how to use it, see the newly added parameters and descriptions. 

3. fancy_multisamp_ideogram - Add functionalities to highlight what's different (or shared) between samples. Adding a new parameter that calls cnvKompare if called correctly. 

4. fancy_circosplot - Circos plot now takes additional variant call sets. Previously it only plotted manta calls, but can now be run so that ssm calls are plotted (similarly to other fancy plotting functions).

5. fancy_alignment_plot - Remove lines connecting the dots.

6. fancy_qc_plot - Barplot has now been replaced with boxplot + jitter for visualizing QC metrics across different sample cohorts, pathologies, etc.

This PR also includes a few minor QoL updates. Including an update to cnvKompare to always retain the same columns in the returns (regardless if concordant or discordant calls are returned) as well as a hot-fix to collate_results (see issue [112](https://github.com/morinlab/GAMBLR/issues/112)).

Before review of this PR, I want to include one more commit I am working on as well as some more thorough testing of the updated plotting functions. I will call for reviewers when the PR is ready to be reviewed

NAMESPACE
```diff
@@ -101,6 +101,7 @@ export(tidy_lymphgen)
 export(web_initialize_gambl_site)
 import(ComplexHeatmap)
 import(DBI)
+import(GenomicRanges)
```
Kdreval, line 104: Can you please give an idea on what the GenomicRanges are used for? If it is just to overlap the coordinates, we can consider using the foverlaps and avoid adding another big-sized dependecy

R/utilities.R
```diff
@@ -842,11 +843,12 @@ maf_to_custom_track = function(maf_data,
 #' @import tidyverse config
 #'
 #' @examples
-#' everything_collated = collate_results(join_with_full_metadata = TRUE)
```
Kdreval, line 845: Can we keep the old name of the argument to maintain backward compatibility?

R/utilities.R
```diff
@@ -842,11 +843,12 @@ maf_to_custom_track = function(maf_data,
...
+#' everything_collated = collate_results(join_with_metadata = TRUE)
 #'
 collate_results = function(sample_table,
                            write_to_file = FALSE,
-                           join_with_full_metadata = FALSE,
+                           join_with_metadata = FALSE,
+                           user_provided_metadata,
```
Kdreval, line 851: We might also name this `these_samples_metadata` similar to other functions, so it is consistent throughout

R/utilities.R
```diff
@@ -889,18 +891,24 @@ collate_results = function(sample_table,
...
+    }
+
+    full_table = left_join(meta_data, sample_table)
+
+    full_table = full_table %>%
+      mutate("MYC_SV_any" = case_when(ashm_MYC > 3 ~ "POS", manta_MYC_sv == "POS" ~ "POS", ICGC_MYC_sv == "POS" ~ "POS", myc_ba == "POS" ~ "POS", TRUE ~ "NEG"))
 
-  full_table = full_table %>%
-    mutate("BCL2_SV_any" = case_when(ashm_BCL2 > 3 ~ "POS", manta_BCL2_sv == "POS" ~ "POS", ICGC_BCL2_sv == "POS" ~ "POS", bcl2_ba == "POS" ~ "POS", TRUE ~ "NEG"))
+    full_table = full_table %>%
```
Kdreval, line 906: Seems like here (and below) just the indentation change?

R/viz.R
```diff
@@ -3710,6 +3716,28 @@ fancy_ideogram = function(this_sample,
     cn_states = mutate(cn_states, chrom = paste0("chr", chrom))
   }
 
+  if(intersect_regions){
+    #filter CN states on intersecting regions
+    #transform regions to Granges objects
+    grl = GRanges(cn_states)
```
Kdreval, line 3722: Alright, see it here now! Can we consider switching this to `data.table::foverlaps()`?

R/viz.R
```diff
@@ -3770,20 +3798,50 @@ fancy_ideogram = function(this_sample,
     maf_trans$ystart = as.integer(maf_trans$ystart)
     maf_trans$yend = as.integer(maf_trans$yend)
 
+    if(intersect_regions){
```
Kdreval, line 3801: Same to this intersect 😃

R/viz.R
```diff
@@ -3794,15 +3852,15 @@ fancy_ideogram = function(this_sample,
...
-    {if(include_ssm) geom_segment(data = maf_ins, aes(x = mid, xend = mid, y = ystart - 0.35, yend = yend - 0.35), color = "black", lineend = "round", size = 3.5, stat = "identity", position = position_dodge())} + #ins
-    {if(include_ssm) geom_segment(data = maf_ins, aes(x = mid, xend = mid, y = ystart - 0.35, yend = yend - 0.35, color = "INS"), lineend = "round", size = 3, stat = "identity", position = position_dodge())} + #ins
-    {if(ssm_count) annotate(geom = "text", x = -4000000, y = del_count$yend, label = del_count$n, color = "#3A8799", size = 3)} + #count del
-    {if(ssm_count) annotate(geom = "text", x = -2300000, y = segment_data$y, label = " | ", color = "black", size = 3)} + #count sep
-    {if(ssm_count) annotate(geom = "text", x = -1000000, y = ins_count$yend, label = ins_count$n, color = "#E6856F", size = 3)} + #countm ins
+    {if(include_ssm && nrow(maf_trans > 0)) geom_segment(data = maf_del, aes(x = mid - 100000, xend = mid + 100000, y = ystart - 0.27, yend = yend - 0.27), color = "#53B1FC", size = 5, stat = "identity", position = position_dodge())} + #del
```
Kdreval, line 3855: Do you want to create here a variable and then apply it in these plots?
For example:
```
my_condition_1 = include_ssm && nrow(maf_trans > 0)
...
{if my_condition_1) ...
```
This way if this ever will need to change or have new parameter added, you will only need to modify the variable assignment 😅

R/viz.R
```diff
@@ -3850,14 +3910,17 @@ fancy_ideogram = function(this_sample,
...
+#'                                        use_augmented_maf = TRUE)
 #'
 fancy_multisamp_ideogram = function(these_sample_ids,
                                     plot_title = "CN Segments Ideogram",
                                     plot_sub = "grch37",
                                     chr_anno_dist = 3,
                                     chr_select = paste0("chr", c(1:22)),
+                                    kompare = FALSE,
+                                    concordance = TRUE,
+                                    cn_thresholds = data.frame(cn_state = c(0:14), low_thresh = c (0:14), high_thresh = c(1:15)),
```
Kdreval, line 3923: I think this will be simplified once the below part converted to use pipe?

R/viz.R
```diff
@@ -4220,57 +4332,75 @@ fancy_circos_plot = function(this_sample,
     n_track = 4
   }
 
-  #get variants
-  svs = get_combined_sv(sample_ids = this_sample, projection = projection)
+  if(manta_calls){
```
Kdreval, line 4335: This probably can be updated, since the argument name implies manta, but the svs are from manta and GRIDSS? Maybe `use_sv_calls`?

R/viz.R
```diff
@@ -4220,57 +4332,75 @@ fancy_circos_plot = function(this_sample,
...
-  #rename variables
-  names(svs_df)[1] = "CHROM_A"
-  names(svs_df)[2] = "START_A"
-  names(svs_df)[3] = "END_A"
-  names(svs_df)[4] = "CHROM_B"
-  names(svs_df)[5] = "START_B"
-  names(svs_df)[6] = "END_B"
-  names(svs_df)[7] = "TYPE"
+    #rename variables
+    names(svs_df)[1] = "CHROM_A"
```
Kdreval, line 4349: How many columns are there in total and what are the original column names? I think I might be missing the step `#split manta_name variable` above. Maybe there is no column renaming required if you use `dplyr::select()`?

If renaming all you can just call this
```
colnames(svs_df) = c("CHROM_A",...,"TYPE")
```
or if there are many columns, this can become
```
colnames(svs_df)[1:7] = c("CHROM_A",...,"TYPE")
```

R/viz.R
```diff
@@ -4603,24 +4736,22 @@ fancy_alignment_plot = function(these_samples,
...
-#'                                    filter_cohort = "DLBCL_LSARP_Trios",
-#'                                    filter_comp_cohort = "BL_Pediatric",
-#'                                    filter_comp_pathology = "BL",
-#'                                    sort_by = "AverageBaseQuality",
-#'                                    plot_data = "AverageBaseQuality",
-#'                                    fill_by = "pathology",
-#'                                    plot_title = "Average Base Quality",
-#'                                    y_axis_lab = "Base Quality",
-#'                                    seq_type = "genome")
+#' fancy_qc_plot_new(filter_pathology = "FL",
```
Kdreval, line 4739: is the name of the function in this example supposed to stay as is (so no _new) ?

NAMESPACE
mattssca: Indeed, the `GenomicRanges` package is used for intersecting two bed-like files. I will revert this to using `foverlaps` as suggested above. Thanks!

R/utilities.R
mattssca: Totally!

R/viz.R
mattssca: Yes, I will do this. Good suggestion!

R/viz.R
mattssca: This has been updated in my latest commit. I decided to streamline the absolute CN conversions here. Also updating the plotting function to handle CN==2 and as a consequence, I have also updated the `subset_cnstates` helper function to retain CN==2 as well (if the new parameter is set to `TRUE`).

R/viz.R
mattssca: Yes, it has been simplified, indeed :)

R/viz.R
mattssca: Good call!

R/viz.R
mattssca: That one slipped through the cracks. Good catch!

R/viz.R
```diff
@@ -3710,6 +3716,26 @@ fancy_ideogram = function(this_sample,
...
+    #transform regions to data tables
+    incoming_cn = as.data.table(cn_states)
+    regions_sub = as.data.table(some_regions)
+
+    #set keys
+    data.table::setkey(incoming_cn, chrom, start, end)
+    data.table::setkey(regions_sub, chrom, start, end)
+
+    #intersect regions
+    intersect = data.table::foverlaps(incoming_cn, regions_sub, nomatch = 0)
```
Kdreval, line 3730: nice addition 😄

Kdreval (approved):

PR #117, opened 2022-08-11
Adding error messages to functions that reads from local files. 

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```

## PR Comments

This PR includes error messages for the following functions;
`get_ssm_by_samples`
`get_ssm_by_sample`
`get_gambl_metadata`
`add_icgc_metadata` (internal function for get_gambl_metadata)
`get_gambl_outcomes` (internal function for get_gambl_metadata)
`get_combined_sv`
`get_manta_sv`
`get_sample_cn_segments`
`get_coding_ssm`
`annotate_ssm_blacklist`
`collate_results`
`assign_cn_to_ssm`
`get_ssm_by_region`
`get_excluded_samples`

The above-mentioned functions have all been tested on remote GAMBLR and all return the added error message if `Sys.setenv(R_CONFIG_ACTIVE = "remote"` has not been specified. 

Functions have also been tested after specifying the config to remote, and the collection of the above-mentioned functions runs the intended way, except for; `get_combined_sv`, `get_sample_cn_segments`, and `get_ssm_by_region`. I am currently looking into why the files needed for these functions are not accessible when running GAMBLR remote with a VPN connection.

Other small updates in this PR include typos in the README as well as removing trailing white spaces.


R/annotation.R
```diff
@@ -50,6 +50,13 @@ annotate_ssm_blacklist = function(mutations_df,
   blacklist_list = list()
   for(b in blacklist_files){
     full_path = paste0(project_base,b)
+
+    #check for missingness
+    if(!file.exists(full_path)){
+      message("Cannot find file locally. If working remotely, perhaps you forgot to load your config (see below) or sync your files?")
```
rdmorin, line 56: Can you also print out the full path before this message i.e. print(full_path) so the user knows what file is missing?

R/database.R
```diff
@@ -16,6 +16,14 @@ maf_header = c("Hugo_Symbol"=1,"Entrez_Gene_Id"=2,"Center"=3,"NCBI_Build"=4,"Chr
 #'
 get_excluded_samples = function(tool_name = "slms-3"){
   base = config::get("repo_base")
+
+  #check for missingness
+  path = paste0(base,"config/exclude.tsv")
+  if(!file.exists(path)){
```
rdmorin, line 22: This is looking in the gambl repo rather than for synced data. Be careful your error messages are not misleading. This should say something along the lines of "have you cloned the GAMBL repo and added the path to this directory under the local section of  your config?" Ideally we would point to some documentation here too.

R/database.R
```diff
@@ -865,6 +925,13 @@ add_icgc_metadata = function(incoming_metadata){
   icgc_publ = mutate(icgc_publ, sex = str_to_upper(sex))
 
   icgc_raw_path = paste0(repo_base,"data/metadata/raw_metadata/ICGC_MALY_seq_md.tsv")
+
```
rdmorin, line 928: The files in this function are mostly in the GAMBL repo. Please refer to my earlier comment and revise these messages accordingly

R/utilities.R
```diff
@@ -127,19 +127,113 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
 #'
 gene_to_region = function(gene_symbol,
```
rdmorin, line 128: This function can no longer return the region in the original chr:start-end format, right?  Why was that functionality removed rather than configuring the function to default to that behaviour unless otherwise specified? I think that restoring the deleted line 139 would actually solve this.

NAMESPACE
```diff
@@ -19,6 +19,7 @@ export(check_host)
 export(cleanup_maf)
 export(cnvKompare)
 export(collate_lymphgen)
+export(collate_qc_results)
```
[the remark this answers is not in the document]
mattssca: Indeed. I have removed the export from this function but have not pushed the updated NAMESPACE just yet.

R/annotation.R
mattssca: Absolutely!

R/utilities.R
```diff
@@ -2741,45 +2849,52 @@ collate_lymphgen = function(sample_table,
...
     stop("Please provide a valid seq_type (\"genome\" or \"capture\").")
   }
-
+  
   #get paths
   base = config::get("project_base") #base
   qc_template = config::get("qc_met") #qc metric template
   qc_path = glue::glue(qc_template) #glue wildcards
   icgc_qc_path_full = paste0(base, qc_path) #combine for icgc_dart qc metric path
   gambl_qc = gsub("icgc_dart", "gambl", qc_path) #use gsub to substitute icgc_dart in qc path with gambl
```
rdmorin, line 2866: This is not the standard way we substitute in wildcards. You should look at the other examples where we use glue.

R/utilities.R
mattssca: If you run this function with the default parameter for return_as, you get the original format chr:start-end. this parameter was added so that you can specify "df" to get regions returned as a data frame in R (for improved compatibility with other functions that can take a df with regions/genes, e.g the new subset parameter in fancy_ideogram).

R/utilities.R
```diff
@@ -2741,45 +2849,52 @@ collate_lymphgen = function(sample_table,
...
     full_qc = rbind(gambl_qc, icgc_qc)
+    sample_table = left_join(sample_table, full_qc)
 
-    return(full_qc)
+    #print n samples with QC metrics
+    qc_samples = length(unique(full_qc$sample_id))
+    message(paste("QC metrics for", qc_samples, "samples retrieved"))
+    
+  }else{
+    message("Currently, seq_type_filter = \"capture\" is only available for unix_group \"icgc_dart\". Only QC metrics for icgc_dart will be returned.")
```
rdmorin, line 2889: I'm confused. This function is implying that capture isn't supported for the other unix group but why?

R/utilities.R
mattssca: Will do!

R/viz.R
```diff
@@ -4828,29 +4861,32 @@ fancy_qc_plot = function(these_samples,
...
+  if(!missing(these_samples_metadata)){
+    these_samples = dplyr::select(these_samples_metadata, sample_id) %>%
+      as.data.frame(strings.as.factors = FALSE) %>%
+      pull(sample_id)
+  }
+
   #filter metadata on selected cohort/pathology
-  if(missing(these_samples)){
+  if(missing(these_samples) && missing(these_samples_metadata)){
     if(!missing(filter_cohort) && missing(filter_pathology)){
```
rdmorin, line 4872: Should these be named "keep_cohort" or something that's more directionally descriptive? I don't know from the name if it means the cohort I want to keep or drop

R/utilities.R
mattssca: When this function was added, we didn't have any QC metrics generated for capture samples with UNIX group gambl. If I remember correctly. Kostia told me this will be added at some point. Thus the function can not retrieve QC data for gambl capture samples, the file does not exists (yet?).

rdmorin (changes requested):
I have left many questions and some comments. Not yet fully reviewed

R/utilities.R
rdmorin: But the default is bed now, isn't it? That's not the same format as chr:start-end

mattssca: My bad, what do we prefer to call this format then?

R/database.R
rdmorin: Easiest way to work out which is which is that there are two base paths:
repo_base (path to your GAMBL repo)
project_base (where the data will actually be sitting)

R/utilities.R
rdmorin: Would help to see what it looks like. Can you show me the format of what's returned when you run with default arguments right now? I have been referring to this format as just "region". Bed is a specific tab-delimited format but a bed-compatable data frame could also reasonably be called "bed".

R/utilities.R
rdmorin: What does this function return when you provide the capture metadata as the samples table and run with seq_type_filter = "capture"?

mattssca:
```
#metadata
> capture_table = get_gambl_metadata(seq_type_filter = "capture") %>% 
    dplyr::select(sample_id, patient_id, biopsy_id, seq_type)
                                                                                                                                                                                                                                                          
#qc
> capture_results = collate_qc_results(sample_table = capture_table, seq_type_filter = "capture")
Currently, seq_type_filter = "capture" is only available for unix_group "icgc_dart". Only QC metrics for icgc_dart will be returned.                                                                                                                      
Joining, by = "sample_id"
QC metrics for 124 samples retrieved

> head(capture_results)
# A tibble: 6 × 18
  sample_id       patient_id biopsy_id  seq_type AverageBaseQuality AverageInsertSize AverageReadLength PairsOnDiffCHR TotalReads TotalUniquelyMap… TotalUnmappedre… TotalDuplicated… ProportionReads… ProportionReads…
  <chr>           <chr>      <chr>      <chr>                 <dbl>             <dbl>             <dbl>          <dbl>      <dbl>             <dbl>            <dbl>            <dbl>            <dbl>            <dbl>
1 02-11616_tumorA 02-11616   S02-1811   capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
2 02-17305_tumorB 02-17305   VS08-18702 capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
3 03-19969_tumorB 03-19969   S04-12701  capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
4 03-23488_tumorA 03-23488   SU03-8870  capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
5 04-11038_tumorA 04-11038   VS04-661   capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
6 04-11038_tumorB 04-11038   BS18-1998  capture                  NA                NA                NA             NA         NA                NA               NA               NA               NA               NA
# … with 4 more variables: MeanCorrectedCoverage <dbl>, ProportionTargetsNoCoverage <dbl>, ProportionCoverage10x <dbl>, ProportionCoverage30x <dbl>
```

mattssca: Just to show that there actually is QC data for (some of) the samples
```
> capture_results[order(capture_results$AverageBaseQuality),]
# A tibble: 2,271 × 18
   sample_id patient_id biopsy_id seq_type AverageBaseQuality AverageInsertSize AverageReadLength PairsOnDiffCHR TotalReads TotalUniquelyMapped TotalUnmappedreads TotalDuplicatedre… ProportionReads… ProportionReads…
   <chr>     <chr>      <chr>     <chr>                 <dbl>             <dbl>             <dbl>          <dbl>      <dbl>               <dbl>              <dbl>              <dbl>            <dbl>            <dbl>
 1 PBL062T   PBL062     PBL062    capture                33.8              97.9               150        1028972   56373556            56219131             154425           34289904             0.61                1
 2 PBL122T   PBL122     PBL122    capture                34.2             331.                150        4852012   66730816            66524931             205885           42018179             0.63                1
 3 PBL118T   PBL118     PBL118    capture                34.3             198.                150        2495055   84819956            84590864             229092           25782507             0.3                 1
 4 PBL068T   PBL068     PBL068    capture                34.5             196.                150        2891320  124278278           124124434             153844           62118572             0.5                 1
 5 PBL106T   PBL106     PBL106    capture                34.5             145.                150        1709854   91945686            91769605             176081           56366033             0.61                1
 6 PBL112T   PBL112     PBL112    capture                34.6             145.                150        1248420   52293838            52204803              89035           25313779             0.48                1
 7 PBL058T   PBL058     PBL058    capture                34.6             272.                150        3840289   81318044            81040974             277070           20259624             0.25                1
 8 PBL021T   PBL021     PBL021    capture                34.6             161.                150        3510318   85804928            85383321             421607           25233184             0.29                1
 9 PBL042T   PBL042     PBL042    capture                34.6             174.                150        2951602   85557642            85302995             254647           21440046             0.25                1
10 PBL123T   PBL123     PBL123    capture                34.7             138.                150         917569   34191782            34027500             164282            7281258             0.21                1
# … with 2,261 more rows, and 4 more variables: MeanCorrectedCoverage <dbl>, ProportionTargetsNoCoverage <dbl>, ProportionCoverage10x <dbl>, ProportionCoverage30x <dbl>
```

R/utilities.R
```diff
@@ -111,14 +111,14 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
 }
 
 
-
-#' Get regions (bed format) from genes.
+#' Get regions from genes.
 #'
 #' @param gene_symbol Gene symbol (e.g BCL2).
```
rdmorin, line 116: Update to indicate that a vector of one or more gene symbols can be provided (same goes for next argument)

R/utilities.R
```diff
@@ -127,19 +127,113 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+    message(paste0(nrow(region), " region(s) returned for ", length(ensembl_id), " gene(s)"))
   }
-  region = paste0(gene_coordinates$chromosome, ":", gene_coordinates$start, "-", gene_coordinates$end)
+
   return(region)
 }
 
+#' Return gennes residing in defined region(s)
+#'
+#' @param region Regions to intersect genes with, this should be a bed-like df (chromosome, start, end).
```
rdmorin, line 198: We should let the user specify either a vector of regions (like the default output of gene_to_region) or this data frame for flexibility. There are some other functions that work the same way (get_ssm_by_region, maybe?)

R/utilities.R
```diff
@@ -127,19 +127,113 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+  if(!str_detect(genes$chromosome, "chr")){
+    genes = mutate(genes, chromosome = paste0("chr", chromosome))}
+
+  genes = genes[genes$chromosome %in% chr_select, ]
+
+  genes = as.data.frame(genes) %>%
+      dplyr::arrange(chromosome, start)
+
+  message(paste0(nrow(genes), " gene(s) returned for ", nrow(region), " region(s)"))
+
```
rdmorin, line 257: Does this return the outer coordinates of each gene in all the regions provided? I can't tell from the documentation

R/utilities.R
mattssca: Yes, I am working on some updates to this function based on the comments regarding gene_to_region (since they are really similar, this one is basically the reverse version of the other function)

R/utilities.R
```diff
@@ -127,19 +127,113 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+#' @export
+#'
+#' @examples
+#' myc_region = gene_to_region(gene_symbol = "MYC", genome_build = "grch37", return_as = "df")
+#' region = region_to_gene(region = myc_region, gene_format = "hugo", genome_build = "grch37")
+#'
+region_to_gene = function(region,
+                          gene_format = "hugo",
+                          genome_build = "grch37",
+                          chr_select = paste0("chr", c(1:22))){
```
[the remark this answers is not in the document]
mattssca: 100% True, Having this parameter in this function is not necessary.

R/utilities.R
mattssca: Yes, together with the regions provided as such;
```
> region_to_gene(region = myc_region, gene_format = "hugo", genome_build = "grch37")
1 gene(s) returned for 1 region(s)
  chromosome     start       end hugo_symbol region_start region_end
1       chr8 128747680 128753674         MYC    128747680  128753674
```

rdmorin: Is it a coincidence that region_start and start are identical and region_end and end are identical? What happens if you specify a larger region as input? What was myc_region here?

mattssca: For this example I first ran gene_to_region and then gave that output to region_to_gene, as such;
```
> myc_region = gene_to_region(gene_symbol = "MYC", genome_build = "grch37", return_as = "df")
1 region(s) returned for 1 gene(s)
> region_to_gene(region = myc_region, gene_format = "hugo", genome_build = "grch37")
1 gene(s) returned for 1 region(s)
  chromosome     start       end hugo_symbol region_start region_end
1       chr8 128747680 128753674         MYC    128747680  128753674
```

Here's an example with a larger region (p-arm chromosome 1);
```
> some_region = data.frame(chromosome = "1", start = 10000, end = 121500000)
> region_to_gene(region = some_region, gene_format = "ensembl", genome_build = "grch37")
2785 gene(s) returned for 1 region(s)
    chromosome   start     end ensembl_gene_id region_start region_end
1         chr1   11869   14412 ENSG00000223972        10000  121500000
2         chr1   14363   29806 ENSG00000227232        10000  121500000
3         chr1   29554   31109 ENSG00000243485        10000  121500000
4         chr1   29554   31109 ENSG00000243485        10000  121500000
5         chr1   29554   31109 ENSG00000243485        10000  121500000
```
(only printed the first 5 rows, there are 2785 genes returned for the specified region)

rdmorin: Why are lines 3, 4 and 5 identical?

mattssca: According to what's in the `grch37_gene_coordinates` R data file, these MIR1302 genes have the same ensemble ID;
```
> grch37_sub = dplyr::filter(grch37_gene_coordinates, chromosome == "1") %>% 
+   arrange(start) %>%
+   head(5) %>%
+   print()
# A tibble: 5 × 6
  ensembl_gene_id chromosome start   end gene_name  hugo_symbol
  <chr>           <chr>      <dbl> <dbl> <chr>      <chr>      
1 ENSG00000223972 1          11869 14412 DDX11L1    DDX11L1    
2 ENSG00000227232 1          14363 29806 WASH7P     WASH7P     
3 ENSG00000243485 1          29554 31109 MIR1302-10 MIR1302-11 
4 ENSG00000243485 1          29554 31109 MIR1302-10 MIR1302-10 
5 ENSG00000243485 1          29554 31109 MIR1302-10 MIR1302-9  
```

the same snippet but with hugo symbols looks like this;
```
> region_to_gene(region = some_region, gene_format = "hugo", genome_build = "grch37")
2785 gene(s) returned for 1 region(s)
    chromosome   start     end hugo_symbol region_start region_end
1         chr1   11869   14412     DDX11L1        10000  121500000
2         chr1   14363   29806      WASH7P        10000  121500000
3         chr1   29554   31109  MIR1302-11        10000  121500000
4         chr1   29554   31109  MIR1302-10        10000  121500000
5         chr1   29554   31109   MIR1302-9        10000  121500000
```

rdmorin: Got it. I think that since what we return doesn't always contain the same columns the data frame we return should be collapsed to unique rows. Does that make sense?

R/utilities.R
```diff
@@ -2731,6 +2906,50 @@ collate_lymphgen = function(sample_table,
...
+#'
+#' @return
+#' @export
+#' @import tidyverse
+#'
+#' @examples
+#' metadata = get_gambl_metadata()
+#' colalted_meta = collate_lymphgen(sample_table = metadata)
+#' 
+collate_lymphgen = function(sample_table){
```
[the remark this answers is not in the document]
lkhilton: Also it would be better to call this argument `these_samples_metadata` for consistency with a lot of other GAMBLR functions. It could also be optional to join it to the metadata table, and could alternatively be provided with a list of sample_ids via `these_sample_ids`.

R/utilities.R
```diff
@@ -2731,6 +2906,50 @@ collate_lymphgen = function(sample_table,
...
+  lymphgen_template = config::get("results_versioned")$lymphgen_template #get tamplate for lymphgen file path
+  lymphgen_path = paste0(repo_base, lymphgen_template) #combine with repo_base path
+  lymphgen_full_paths = glue::glue(lymphgen_path, flavour = flavours) #substitute {flavour} wildcard in template with actual flavours
+  
+  #print paths
+  message(paste0("Reading from the following paths:"))
+  message(paste0(for(i in lymphgen_full_paths) print(i)))
+  
+  #read files into R
+  for (f in lymphgen_full_paths){
```
lkhilton, line 2937: I can think of a few edge cases where this function could break, namely whether any new versions of LymphGen outputs (i.e. hypothetical future files that don't start with `GAMBL_all_the_things`) don't have all of the "flavour" combinations. There's also a ton of useful information in the LymphGen table besides just the classification, and there might be use cases to use those as well. 

This is how I would iterate over this: 
```
load_lymphgen <- function(flavour, lymphgen_path){
    lg_path <- glue::glue(lymphgen_path, flavour = flavour)
    if(!file.exists(lg_path){return()} # Ignore missing flavours
    lg_df <- read_tsv(lg_path) %>% 
        mutate(flavour = flavour) # Append the flavour in its own column called "flavour"
    return(lg_df)
}
lymphgen_results <- lapply(flavours, load_lymphgen, lymphgen_path = lymphgen_path)
lymphgen_results <- bind_rows(lymphgen_paths) # Now you have all the lymphgen results tables stacked on top of each other, with the results from each flavour identified by the `flavour` column
lymphgen_results <- tidy_lymphgen(lymphgen_results, lymphgen_column_in = "Subtype.Prediction", lymphgen_column_out = "LymphGen") 
if(!tidy){ 
    # "Tidy" results are usually long format, so the wide format is the "untidy" version
    lymphgen_untidy <- lymphgen_results %>% 
        select(Sample.Name, Subtype.Prediction, LymphGen, flavour) %>% 
        # This keeps both the original Subtype.Prediction and tidied LymphGen values and puts the values from each "flavour" in its own column
        pivot_wider(
            names_from = flavour, 
            values_from = c(Subtype.Prediction, LymphGen), 
            names_glue = c({.value}_{flavour})
        )
    # Do some stuff to join this with the metadata (if provided) or subset to the sample_ids (if provided)
    ...
    return(lymphgen_untidy)
} else {
    # Join with metadata or subset to sample_ids
    return(lymphgen_tidy)
    # Notice I didn't drop any columns from this version. There's lots of information in the "tidy" table about feature_count, list of features used in classification, classification confidence, etc. that can be kept in the tidy version. 
}

```

config.yml
```diff
@@ -16,14 +16,14 @@ default:
...
         ex_matrix_file: "/projects/rmorin/projects/gambl-repos/gambl-rmorin/results/icgc_dart/DESeq2-0.0_salmon-1.0/mrna--gambl-icgc-all/vst-matrix.tsv"
-        lymphgen_template: "all_the_things/lymphgen-1.0/99-outputs/GAMBL_all_the_things.lymphgen_calls.{flavour}.tsv"
         deblacklisted: "all_the_things/slms_3-1.0_vcf2maf-1.3/{seq_type}--projection/deblacklisted/maf/all_slms-3--{projection}.CDS.maf"
         augmented: "all_the_things/slms_3-1.0_vcf2maf-1.3/{seq_type}--projection/deblacklisted/augmented_maf/all_slms-3--{projection}.CDS.maf"
         cnv: "all_the_things/cnv_master-1.0/merges/{seq_type}--{projection}.seg"
         sv: "all_the_things/svar_master-1.0/merges/gridss_manta.genome--{projection}.bedpe"
     results_merged_wildcards:
         lymphgen_template: "no_cnvs.no_sv.with_A53,no_cnvs.no_sv.no_A53,no_cnvs.with_sv.with_A53,with_cnvs.no_sv.no_A53,no_cnvs.with_sv.no_A53,with_cnvs.no_sv.with_A53,with_cnvs.with_sv.no_A53,with_cnvs.with_sv.with_A53"
     results_versioned:
+        lymphgen_template: "versioned_results/LymphGen/GAMBL_all_the_things.lymphgen_calls.{flavour}.tsv"
```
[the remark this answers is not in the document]
rdmorin: Aren't we also going to rename LymphGen to match the directory name under results/gambl, possibly as shown in the new_version example?

R/utilities.R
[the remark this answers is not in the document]
rdmorin: What if we made these_samples_metadata optional and had the function return the rows for all samples in the files by default?

R/utilities.R
```diff
@@ -2731,6 +2906,50 @@ collate_lymphgen = function(sample_table,
...
+collate_lymphgen = function(sample_table){
+  
+  #setup paths
+  repo_base = config::get("repo_base") #get repo base
+  flavours = config::get("results_merged_wildcards")$lymphgen_template #get all flavours
+  flavours = str_split(flavours, pattern = ",") #seperate each flavour on ","
+  flavours = unlist(flavours) #unlist flavours
+  lymphgen_template = config::get("results_versioned")$lymphgen_template #get tamplate for lymphgen file path
+  lymphgen_path = paste0(repo_base, lymphgen_template) #combine with repo_base path
+  lymphgen_full_paths = glue::glue(lymphgen_path, flavour = flavours) #substitute {flavour} wildcard in template with actual flavours
```
rdmorin, line 2930: Does this actually work? I thought it would just substitute in a vector from namespace so the part specifying flavour=flavours isn't necessary if you instead had your vector named "flavour".

R/utilities.R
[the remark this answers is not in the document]
lkhilton: Agree, Ryan,  this would also be consistent with a lot of other GAMBLR functions.

config.yml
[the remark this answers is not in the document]
lkhilton: Yes, I need to fix that

R/utilities.R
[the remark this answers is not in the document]
rdmorin: Sort of. Many GAMBLR functions default to calling get_gambl_metadata under the hood. In this scenario, I'm saying the function instead rely on the sample IDs in the files it's loading (ignoring the metadata entirely)

config.yml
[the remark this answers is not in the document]
mattssca: I have updated the config based on the comments here. I also left an outstanding TODO, that the key needs to be updated once the renaming has taken place.

R/utilities.R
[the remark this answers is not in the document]
mattssca: `these_samples_metadata` has been added as an optional parameter (replacing sample_table) if provided, lymphgen will be joined with this data frame. If not provided, the function will return the rows for all the samples in the files (default).

R/utilities.R
mattssca: Great suggestion, I have incorporated parts of this into the updated function, thanks @lkhilton

R/utilities.R
```diff
@@ -2908,48 +2908,76 @@ consolidate_lymphgen = function(sample_table,
...
-    this_data = dplyr::select(this_data, Sample.Name, flavour)
-    colnames(this_data)[1] = "sample_id" #change name of first column to match sample_table
-    sample_table = sample_table %>% #left join with incoming sample_table (by sample_id)
-      left_join(this_data, by = "sample_id")
+#' this_meta = get_gambl_metadata() %>% dplyr::filter(pathology == "FL")
+#' wide_lymphgen = collate_lymphgen(these_samples_metadata = this_meta, lymphgen_version = "default", tidy_lymphgen = FALSE)
+#'
+collate_lymphgen = function(these_samples_metadata,
+                            lymphgen_version = "default",
+                            tidy_lymphgen = FALSE){
```
lkhilton, line 2925: I can think of at least one other function where the option is just `tidy` and that controls whether the output is long or wide, so could you just call this `tidy`? (Also so it's not easily confused with the `tidy_lymphgen` function).

R/utilities.R
```diff
@@ -2908,48 +2908,76 @@ consolidate_lymphgen = function(sample_table,
...
-  
-  #read files into R
-  for (f in lymphgen_full_paths){
-    this_data = fread(f) #read with fread
-    this_data = tidy_lymphgen(this_data, lymphgen_column_in = "Subtype.Prediction", lymphgen_column_out = "flavour") 
-    this_data = dplyr::select(this_data, Sample.Name, flavour)
-    colnames(this_data)[1] = "sample_id" #change name of first column to match sample_table
-    sample_table = sample_table %>% #left join with incoming sample_table (by sample_id)
-      left_join(this_data, by = "sample_id")
+#' this_meta = get_gambl_metadata() %>% dplyr::filter(pathology == "FL")
```
lkhilton, line 2920: Maybe use "DLBCL" as the pathology in this example because LymphGen is predominantly designed for DLBCL.

R/utilities.R
```diff
@@ -2908,48 +2908,76 @@ consolidate_lymphgen = function(sample_table,
...
+      if(!missing(these_samples_metadata)){
+        meta_data = these_samples_metadata
+        lymphgen_untidy = left_join(meta_data, lymphgen_untidy, by = c("sample_id" = "Sample.Name"))
+      }
+
+      return(lymphgen_untidy)
+
+    }else{
+    if(!missing(these_samples_metadata)){
+      meta_data = these_samples_metadata
```
lkhilton, line 2971: Seems an unnecessary step? Why not just join `these_samples_metadata` with the lymphgen table directly?

R/utilities.R
```diff
@@ -2908,48 +2908,76 @@ consolidate_lymphgen = function(sample_table,
...
+      lymphgen_results = left_join(meta_data, lymphgen_results, by = c("sample_id" = "Sample.Name"))
+    }
+
+    return(lymphgen_results)
```
lkhilton, line 2975: Could you please make it so that any table returned by this function uses `sample_id` instead of `Sample.Name`? This will make it easier to work with for downstream filtering steps. You could implement the change in the `load_lymphgen` function.

lkhilton (commented):
Looks :delicioso: Adam!

R/utilities.R
mattssca: Yes, makes sense!

R/utilities.R
mattssca: Indeed, an unnecessary step. Good catch :)

lkhilton (approved):
The collate_lymphgen function looks good to me - thanks for addressing my comments!

R/database.R
```diff
@@ -227,6 +227,20 @@ get_ssm_by_samples = function(these_sample_ids,
           verbose = FALSE,
           ssh_session = ssh_session
         )
+
```
lkhilton, line 230: I think this method would throw an error if the user opted for `basic_columns = FALSE` since this doesn't specify `col_types` for columns beyond the first ~45. 

I think the place to set this is in the `fread_maf` function. It already has a few `col_types` specified explicitly, but it would need to be extended to cover all columns in case of an empty or near-empty file. `fread_maf` uses `fread`, @rdmorin should it be updated to use `read_tsv`? One of the handy things about `read_tsv` is that you can load the column specifications for a good maf file, save that as an R object, then use that to specify the column types in a `read_tsv` function. There's a useful example [here](https://readr.tidyverse.org/reference/spec.html#ref-examples).

lkhilton (changes requested):

R/database.R
mattssca: Thanks for your comment. fread_maf has now been updated (and get_ssm_by_samples reverted) to hold column types for all 116 maf-columns.

R/preprocessing_io.R
```diff
@@ -750,7 +750,7 @@ fread_maf = function(maf_file_path){
...
   return(maf_dt)
```
lkhilton, line 754: For readability and editability, it's much better to break up long lists of parameters onto multiple lines. In fact the R style guide says one parameter per line, and no lines should be longer than 80 characters. Some style tips here: https://style.tidyverse.org/syntax.html#syntax

I tried loading in one of the CDS mafs with read_tsv, and the col_types for several of the columns don't match what you have here. You can check it with the following command: 
```
spec(read_tsv("results/all_the_things/slms_3-1.0_vcf2maf-1.3/capture--projection/deblacklisted/maf/all_slms-3--grch37.CDS.maf"))
```

If we updated this function to use `read_tsv` instead of `fread`, you could do this: 
```
# Get a minimal specification of col_types by reading in the CDS maf file 
maf_col_types <- cols_condense(spec(read_tsv("results/all_the_things/slms_3-1.0_vcf2maf-1.3/capture--projection/deblacklisted/maf/all_slms-3--grch37.CDS.maf")))
# Save this as an object to the data directory
save(maf_col_types, file = "data/maf_col_types.rda")

# In all future instances of fread_maf or any other maf-loading function (including `get_ssm_by_samples(subset_from_merge = TRUE)`, you can use this object
maf_col_types <- load("data/maf_col_types.rda")
fread_maf <- function(maf_file_path){
  maf_dt <- readr::read_tsv(
    maf_file_path, 
    ... 
    col_types = maf_col_types
  )
)
}
```

lkhilton (changes requested):

R/preprocessing_io.R
rdmorin: Any creation of an rda file like this should happen in this R script for reproducibility:
data-raw/DATASET.R

rdmorin: Adam you should familiarize yourself with that script since it tracks how every rda file (shared object) in GAMBLR came to be. I would add that any file that script uses should ALSO exist in the repo (you can see examples in that script). Hence, any demo MAF snippit really should be put into the repo if it will be used by this script.

mattssca: Thanks for your comment @lkhilton. Yes, I thought about this yesterday too and I think the reason that I got some of the column types wrong was because I took a random maf and checked the column types of it and all columns that had NA most likely got interpreted as logical. It does not seem to be a problem with fread, since the function only overrides coltypes of a higher type. But nonetheless, the column types specified here should match the expected type. I will update the column types as well as improve the readability.

mattssca: Sounds good @rdmorin I'll go over this script. Thanks,

lkhilton: Ryan and I debated this and I think this is an instance where generating this .rda object doesn't necessarily have to be done through the DATASET.R script, and I don't think it's worth adding a maf snippet to this repository just to be able to reproduce that object. It's still good for you to see how the DATASET.R script works though.

mattssca: I have now updated the column classes that were not a match. Mostly, columns that wrongly got classified as logical because of missing values in the template MAF I used to interrogate column classes. I have also improved the readability as suggested. In my next PR, I can work towards switching from fread to read_tsv as detailed above, if we think this is a good strategy.

rdmorin:
Is this ready for re-review?

mattssca:
Yes, it's ready.

rdmorin:
OK. Can you also please respond to my question above?

R/preprocessing_io.R
```diff
@@ -733,24 +733,156 @@ find_files_extract_wildcards = function(tool_results_path,
 
 #' Title
 #'
-#' @param maf_file_path
+#' @param maf_file_path path to maf that is about to be read.
+#' @param select_cols Optional parameter for specifying what columns are to be returned. If not specified, all columns will be kept.
```
rdmorin, line 737: Can you describe (is it a numeric vector? Character vector of column names? Either?)

rdmorin (changes requested):
See individual comments. Just wondering about a few changes

R/database.R
```diff
@@ -196,23 +191,32 @@ get_ssm_by_samples = function(these_sample_ids,
       maf_path = glue::glue(maf_template)
       full_maf_path =  paste0(config::get("project_base"), maf_path)
       message(paste("using existing merge:", full_maf_path))
-
-      #check for missingness
-      if(!file.exists(full_maf_path)){
```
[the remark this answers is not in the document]
mattssca: Not sure, maybe this was an unfortunate case of auto-merge. I'll add it back in.

mattssca:
This PR has now been updated based on the latest review comments, in [this](https://github.com/morinlab/GAMBLR/commit/c185f36dd3e100e0092ee6d0815cc1c45797ac31) commit.

R/utilities.R
mattssca: This has been implemented.

R/preprocessing_io.R
rdmorin: Sorry my question was actually about `select_cols Optional parameter `

mattssca: Ah ok I see. Let's update that one as well.

lkhilton (approved):

PR #129, opened 2022-10-14
update lymphoma genes

rdmorin:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


rdmorin:
This addresses some long-standing bugs that may also have been fixed in Adam's branch. If this gets to Master first, we'll need to be careful about not reversing/breaking some of these changes when that branch eventually gets merged. I added a new R script that is meant as a basis for testing that all our function's examples actually work. That code could be used to make some of the new Vignettes.

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
...
+  #drop genes not in the list then tally only coding variants (by default). 
+  # TODO: eventually allow an option to collapse samples from the same patient
+  if(missing(other_genes)){
+    these_genes_maf = dplyr::filter(maf_df,Hugo_Symbol %in% these_genes)
+  }else{
+    these_genes_maf = dplyr::filter(maf_df,Hugo_Symbol %in% c(these_genes,other_genes))
+  }
+  
+  #drop non-coding
+  these_genes_maf = dplyr::filter(these_genes_maf,Variant_Classification %in% coding_vc)
```
Kdreval, line 405: Is the `coding_vc` defined globally in GAMBLR?

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
...
+#' @param colour_index 
+#'
+#' @return data frame with counts for each gene
+#' @export
+#'
+#' @examples
+prettyGeneCloud = function(maf_df,these_genes,other_genes,
+                           these_genes_colour="#B2DF8A",
+                           other_genes_colour="#bc42f5",
+                           colour_index){
```
Kdreval, line 392: Is the color_index here a named list where each gene has a corresponding unique color?

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
...
+    if(!missing(colour_index)){
+      #use the colours in colour_index to colour the gene names
+      if(any(!these_genes %in% names(colour_index))){
+        stop("all genes in these_genes must be among the names of colour_index if you specify this variable")
+      }
+      these_genes_unique$color=colour_index[these_genes_unique$Hugo_Symbol]
+      #make cloud with the user-specified colours mapped to the genes
+      wordcloud(these_genes_unique$Hugo_Symbol,these_genes_unique$n,random.order=F,ordered.colors=T,colors=these_genes_unique$color)
+    }else{
+      wordcloud(these_genes_unique$Hugo_Symbol,these_genes_unique$n,random.color=TRUE,colors=brewer.pal(12,"Set3"))
```
Kdreval, line 425: I don't think RColorBrewer is GAMBLR dependency either, so this might error out if not installed/loaded

R/database.R
```diff
@@ -160,24 +188,56 @@ get_ssm_by_samples = function(these_sample_ids,
...
-      maf_df_merge = fread_maf(full_maf_path) %>%
-        dplyr::filter(Tumor_Sample_Barcode %in% these_sample_ids)
-      #subset maf to only include first 43 columns (default)
-      if(basic_columns){maf_df_merge = dplyr::select(maf_df_merge, c(1:45))}
-      #subset maf to a specific set of columns (defined in maf_cols)
-      if(!is.null(maf_cols) && !basic_columns){maf_df_merge = dplyr::select(maf_df_merge, all_of(maf_cols))}
+      if(engine=="fread_maf"){
+        if(basic_columns){
+          maf_df_merge = fread_maf(full_maf_path,select_cols = c(1:45)) %>%
+            dplyr::filter(Tumor_Sample_Barcode %in% these_sample_ids) %>%
```
Kdreval, line 194: Do you want to put this and the next line outside the ifelse check, since both filters are the same?
EDITED:: It can also probably be it's own non-exported small function, since it is also used below for the augmented variants

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
...
 
+gene_mutation_tally = function(maf_df,these_samples_metadata,these_genes,grouping_variable="cohort"){
+  meta = dplyr::select(these_samples_metadata,sample_id,{{grouping_variable}})
+  maf_filt = dplyr::filter(maf_df,Hugo_Symbol %in% these_genes, Variant_Classification %in% coding_class) %>% 
+    dplyr::filter(Variant_Classification !="Silent")
+  meta_anno = left_join(maf_filt,meta,by=c("Tumor_Sample_Barcode"="sample_id")) %>% 
+    group_by(Hugo_Symbol,Tumor_Sample_Barcode) %>% 
+    slice_head() %>% 
+    ungroup()
+  denom = meta %>% group_by(!!sym(grouping_variable)) %>% tally() %>% dplyr::rename(c("total"="n"))
```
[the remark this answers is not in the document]
rdmorin: Doesn't mutate leave redundant rows? Maybe it should use `summarise()` though

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
   return(p)
 }
 
+gene_mutation_tally = function(maf_df,these_samples_metadata,these_genes,grouping_variable="cohort"){
```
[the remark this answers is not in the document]
rdmorin: It's a helper function for a plotting function though and not exported. It probably should have documentation added but I don't see it as pressing currently

R/viz.R
```diff
@@ -359,6 +359,78 @@ prettyRainfallPlot = function(this_sample_id,
...
+
+#' Make a word cloud of gene names from a MAF file based on mutation frequency
+#'
+#' @param maf_df 
+#' @param these_genes 
+#' @param other_genes 
+#' @param these_genes_colour 
+#' @param other_genes_colour 
+#' @param colour_index 
+#'
```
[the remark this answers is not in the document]
rdmorin: Should be fixed now

R/viz.R
rdmorin: Documented

R/viz.R
rdmorin: I believe so, yes. Probably in the first few lines of database.R or utilities.R

R/database.R
rdmorin: The better approach is actually to have these passed to fread_maf so it can lazily drop the non-matching lines on the fly rather than storing them in memory only to have them dropped a few lines later. I think Adam can work on this as a refinement.

Kdreval (approved):

rdmorin:
Trying to deal with inability to resolve conflict

PR #133, opened 2022-10-21
collate_results fix

mattssca:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1286,13 +1286,13 @@ test_glue = function(placeholder="INSERTED"){
 #' Bring together all derived sample-level results from many GAMBL pipelines.
 #'
 #' @param sample_table A data frame with sample_id as the first column.
-#' @param write_to_file Boolean statement that outputs tsv file if TRUE, default is FALSE.
+#' @param write_to_file Boolean statement that outputs tsv file (/projects/nhl_meta_analysis_scratch/gambl/results_local/shared/gambl_{seq_type_filter}_results.tsv) if TRUE, default is FALSE.
 #' @param join_with_full_metadata Join with all columns of meta data, default is FALSE.
 #' @param these_samples_metadata Optional argument to use a user specified metadata df, overwrites get_gambl_metadata in join_with_full_metadata.
 #' @param case_set Optional short name for a pre-defined set of cases.
 #' @param sbs_manipulation Optional variable for transforming sbs values (e.g log, scale).
 #' @param seq_type_filter Filtering criteria, default is genomes.
-#' @param from_cache Boolean variable for using cached results, default is TRUE. If write_to_file is TRUE, this aprameter auto-defaults to FALSE. 
+#' @param from_cache Boolean variable for using cached results (/projects/nhl_meta_analysis_scratch/gambl/results_local/shared/gambl_{seq_type_filter}_results.tsv), default is TRUE. If write_to_file is TRUE, this parameter auto-defaults to FALSE.
 #'
 #' @return A table keyed on biopsy_id that contains a bunch of per-sample results from GAMBL
 #' @export
@@ -1302,10 +1302,10 @@ test_glue = function(placeholder="INSERTED"){
 #' #generate new cached results for all genome samples in gambl (Warning, is this what you really want to do?)
 #' genome_collated = collate_results(seq_type_filter = "genome", from_cache = FALSE, write_to_file = TRUE)
 #'
-#' #get collated results for all capture samples, using chached results
-#' capture_collated_eerything = collate_results( seq_type_filter = "capture", from_cache = TRUE, write_to_file = FALSE)
+#' #get collated results for all capture samples, using cached results
+#' capture_collated_everything = collate_results( seq_type_filter = "capture", from_cache = TRUE, write_to_file = FALSE)
 #'
-#' #use an already subset metadata table for getting collated results (cheched)
+#' #use an already subset metadata table for getting collated results (cached)
 #' metadata = get_gambl_metadata(seq_type_filter = "genome") %>% dplyr::filter(pathology == "FL")
 #' fl_collated = collate_results(seq_type_filter = "genome", join_with_full_metadata = TRUE, these_samples_metadata = metadata, write_to_file = FALSE, from_cache = TRUE)
 #'
@@ -1352,7 +1352,7 @@ collate_results = function(sample_table,
       message('Sys.setenv(R_CONFIG_ACTIVE = "remote")')
     }
 
-    #read chached results
+    #read cached results
     sample_table = read_tsv(output_file) %>% dplyr::filter(sample_id %in% sample_table$sample_id)
 
   }else{
@@ -1369,7 +1369,7 @@ collate_results = function(sample_table,
     sample_table = collate_qc_results(sample_table = sample_table, seq_type_filter = seq_type_filter)
   }
   if(write_to_file){
-    #write results from "slow option" to new chached results file
+    #write results from "slow option" to new cached results file
     write_tsv(sample_table, file = output_file)
   }
   #convenience columns bringing together related information
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -1311,7 +1311,11 @@ test_glue = function(placeholder="INSERTED"){
 #'
 #' #get collated results for all genome samples and join with full metadata
 #' everything_collated = collate_results(seq_type_filter = "genome", from_cache = TRUE, join_with_full_metadata = TRUE)
-
+#'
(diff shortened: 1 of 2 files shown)


R/utilities.R
```diff
@@ -1320,14 +1320,10 @@ collate_results = function(sample_table,
   if(write_to_file){
     from_cache = FALSE #override default automatically for nonsense combination of options
   }
-  output_file = config::get("results_merged")$collated
```
rdmorin, line 1323: This is a partial fix but how will this work when use_cache = TRUE?

rdmorin: I know you said it's only fixing one of the two modes but this code change is guaranteed to not allow the other mode to ever work due to the scope of the variables.

rdmorin (changes requested):
The solution is guaranteed to never work for the other (more common) use case. Please adjust the scope of the variables that are required by both modes to address this.

R/utilities.R
mattssca: So I guess both options (from_cache = TRUE/FALSE) would work if they have access to the same output_files? i.e moving the output_file code outside the from_cache if statement. Allowing from_cache = FALSE to find the file if write_to_file = TRUE. Is this correct?

```
 output_file = config::get("results_merged")$collated
 output_base = config::get("project_base")
 output_file = paste0(output_base, output_file)
 output_file = glue::glue(output_file)
```

rdmorin:
Please don't submit a PR until the problem is fully fixed. Use a draft PR instead if you want feedback on the partial fix. I also would like to see working examples that test both modes of from_cache in the PR.

mattssca:
I have now marked this PR as a draft, I will update and provide the information you requested once it's ready for review. Thanks!

R/utilities.R
```diff
@@ -1320,17 +1320,14 @@ collate_results = function(sample_table,
   if(write_to_file){
     from_cache = FALSE #override default automatically for nonsense combination of options
   }
+  
   output_file = config::get("results_merged")$collated
```
rdmorin, line 1324: Adam after you left we discussed this further and opted to save the results to that `results/shared` folder so please change this line to 
```
output_base = config::get("project_base")
```

rdmorin: In this case those both point to the same file because every gambl clone has a symlink `results` that points to `results_local`. You should familiarize yourself with the layout and organization of the gambl repo especially how we use symlinks. GAMBLR should be using the first path in this case because it's writing to `results` rather than one of the paths in the repo that is tracked in git.

mattssca: Perfect, Thanks for clarifying, makes 100% sense.

mattssca:
This function has now been updated with appropriate file paths for reading and exporting cached results. The function has also been getting a facelift in terms of expanded in-line comments, parameter descriptions and extended examples. 

When testing this function I was not allowed to write to file:  `Error: Cannot open file for writing`

However, I tested this function with `from_cached = FALSE` and `write_to_file = FALSE` for both `genome` and `capture` with success. As such:

```
genome_collated = collate_results(seq_type_filter = "genome", from_cache = FALSE, write_to_file = FALSE)
capture_collated = collate_results(seq_type_filter = "capture", from_cache = FALSE, write_to_file = FALSE)
```
I also tested using `these_samples_metadata` and `join_with_full_metadata`:

```
fl_metadata = get_gambl_metadata(seq_type_filter = "genome") %>% dplyr::filter(pathology == "FL")
fl_collated = collate_results(seq_type_filter = "genome", join_with_full_metadata = TRUE, these_samples_metadata = fl_metadata, write_to_file = FALSE, from_cache = FALSE)

```
Lastly, I tested this function with using the sample_table parameter as well:
```
fl_samples = get_gambl_metadata(seq_type_filter = "genome") %>% dplyr::filter(pathology == "FL") %>% dplyr::select(sample_id, patient_id, biopsy_id)
fl_collated = collate_results(sample_table = fl_samples, seq_type_filter = "genome", from_cache = FALSE, write_to_file = FALSE)
```

R/utilities.R
```diff
@@ -1376,7 +1380,7 @@ collate_results = function(sample_table,
       meta_data = get_gambl_metadata(seq_type_filter = seq_type_filter)
     }
 
-    full_table = left_join(sample_table, meta_data)
+    full_table = left_join(meta_data, sample_table)
```
rdmorin, line 1383: Out of interest, why was this change made?

mattssca: I reverted a change that I made in [this](https://github.com/morinlab/GAMBLR/pull/133/commits/213e9f8f0e3721fefbf3e0de9164173f96d0adfd) commit, from when I first started looking into this function last week.

rdmorin: Then i have the same question but in regards to that commit instead. Why was it changed?

mattssca: I thought it should be the other way around (i.e sample_table , meta_data) but after some thinking and testing, it was correct from the beginning (i.e meta_data, sample_table).

R/utilities.R
```diff
@@ -1292,14 +1292,30 @@ test_glue = function(placeholder="INSERTED"){
 #' @param case_set Optional short name for a pre-defined set of cases.
 #' @param sbs_manipulation Optional variable for transforming sbs values (e.g log, scale).
 #' @param seq_type_filter Filtering criteria, default is genomes.
-#' @param from_cache Boolean variable for using cached results, default is TRUE.
+#' @param from_cache Boolean variable for using cached results, default is TRUE. If write_to_file is TRUE, this aprameter auto-defaults to FALSE.
```
rdmorin, line 1295: There's a typo here (parameter is misspelled)

R/utilities.R
```diff
@@ -1292,14 +1292,30 @@ test_glue = function(placeholder="INSERTED"){
...
 #' @return A table keyed on biopsy_id that contains a bunch of per-sample results from GAMBL
 #' @export
 #' @import tidyverse config
 #'
 #' @examples
-#' everything_collated = collate_results(join_with_full_metadata = TRUE)
+#' #generate new cached results for all genome samples in gambl (Warning, is this what you really want to do?)
+#' genome_collated = collate_results(seq_type_filter = "genome", from_cache = FALSE, write_to_file = TRUE)
+#'
+#' #get collated results for all capture samples, using chached results
```
rdmorin, line 1305: typo (cached)

R/utilities.R
```diff
@@ -1292,14 +1292,30 @@ test_glue = function(placeholder="INSERTED"){
...
+#' capture_collated_eerything = collate_results( seq_type_filter = "capture", from_cache = TRUE, write_to_file = FALSE)
```
rdmorin, line 1306: Presumably you meant to name your variable `capture_collated_eerything`

R/utilities.R
```diff
@@ -1292,14 +1292,30 @@ test_glue = function(placeholder="INSERTED"){
...
+#'
+#' #use an already subset metadata table for getting collated results (cheched)
```
rdmorin, line 1308: cheched?

R/utilities.R
```diff
@@ -1292,14 +1292,30 @@ test_glue = function(placeholder="INSERTED"){
 #' @param case_set Optional short name for a pre-defined set of cases.
 #' @param sbs_manipulation Optional variable for transforming sbs values (e.g log, scale).
 #' @param seq_type_filter Filtering criteria, default is genomes.
-#' @param from_cache Boolean variable for using cached results, default is TRUE.
+#' @param from_cache Boolean variable for using cached results, default is TRUE. If write_to_file is TRUE, this aprameter auto-defaults to FALSE. 
 #'
 #' @return A table keyed on biopsy_id that contains a bunch of per-sample results from GAMBL
```
rdmorin, line 1297: For the sake of our future sanity can you add some details in this section that tell the user where the cached results are written?

R/utilities.R
```diff
@@ -1354,7 +1369,7 @@ collate_results = function(sample_table,
     sample_table = collate_qc_results(sample_table = sample_table, seq_type_filter = seq_type_filter)
   }
   if(write_to_file){
-    
+    #write results from "slow option" to new chached results file
```
rdmorin, line 1372: typo

rdmorin (changes requested):
There are lots of aesthetic fixes that should be applied before this gets merged

rdmorin (commented):
Good to go. Try to be more careful reviewing your documentation for typos before you submit a PR.

rdmorin (approved):

PR #142, opened 2022-11-22
GAMBLR hotfixes and improvements.

mattssca:
This PR includes a variety of hotfixes and updates to existing functions. Including;

- `assign_cn_to_ssm` now takes `this_seq_type` parameter.
- `get_cn_segments` now has support for flat files.
- `prettyOncoplot` can now deal with silent mutations.
- More examples added to `test_functions.R`.
- Improved instructions on how to run remote (README).
- Supressing annoying warning messages from `readr`.
- Updates to `fancy_v_count` so that it correctly calls `assign_cn_to_ssm`.
- Updates to the config (redundant line from `get_gene_expression` update).
- Hotfix for `fancy_x_plot` functions that threw unnecessary writing messages.
- Adding SSM thematic vignette.
- Improvements to `ashm_rainbow_plot` (fixing fill, margins, axis titles, ticks, etc.).
- Added new thematic vignette (CN/SV), in progress...

For more info, please see the commit messages below.

This branch was recently merged with the current master (22/11). Many merge conflicts were flagged, and even though the majority of these conflicts could be resolved automatically, a handful had to be resolved manually. 

I have also tested this branch using the examples available in `test_functions.R`. The functions listed in this script runs without errors and are producing the expected outputs. With that being said, I would highly recommend at least one additional tester to also confirm this given the number of merge conflicts and updates to commonly used GAMBLR functions, and perhaps other examples that are not (yet) represented in test_functions.R

After this PR has been approved, I will close the corresponding issues on this repo.

If this PR should still be active upon my return from Sweden, I have some additional improvements that I am currently working on. But these could also very well be on a future PR.

Let me know if there are any questions. 

Thanks,


R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
-#' @param qend End coordinate of the range you are restricting to.
-#' @param region Region formatted like chrX:1234-5678 instead of specifying chromosome, start and end separately.
-#' @param with_chr_prefix Prepend all chromosome names with chr (required by some downstream analyses).
+#' @param region Region formatted like chrX:1234-5678 or X:1234-56789.
+#' @param chromosome The chromosome you are restricting to. Required parameter if region is not specified.
+#' @param qstart Start coordinate of the range you are restricting to. Required parameter if region is not specified.
+#' @param qend End coordinate of the range you are restricting to. Required parameter if region is not specified.
+#' @param projection Selected genome projection for returned Cn segments.
+#' @param this_seq_type Seq type for returned Cn segments. Currently, only genome is supported. Capture samples will be added once processed through CNV protocols.
+#' @param remove_chr_prefix Prepend all chromosome names with chr (required by some downstream analyses). Default is FALSE.
```
Kdreval, line 1724: Maybe the description should be not "Prepend ...", but rather "Remove ..."? Otherwise does not sound to match the parameter name

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
-#' my_segments = get_cn_segments(chromosome="8",qstart=128723128,qend=128774067)
-#' #Asking for chromosome names to have a chr prefix (default is un-prefixed)
-#' prefixed_segments = get_cn_segments(chromosome ="12",qstart = 122456912, qend = 122464036, with_chr_prefix = TRUE)
-#'
-get_cn_segments = function(chromosome = "",
+#' #Example using chromosome, qstart and qend parameters:
+#' segments_region_grch37 = get_cn_segments(chromosome = "chr8",
+#'                                          qstart = 128723128,
+#'                                          qend = 128774067,
+#'                                          projection = "grch37",
```
Kdreval, line 1738: Projection and genome here are same as default, so maybe let's drop them from example to show the "minimum usage"? The next example then shows how to modify different projection and I think the flatfiles are not available for capture anyways

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
-  table_name = config::get("results_tables")$copy_number
-  table_name_unmatched = config::get("results_tables")$copy_number_unmatched
+  
+  #check seq type and return a message if anything besides "genome" is called. To be updated once capture samples have been processed through CNV protocols.
+  if(this_seq_type!="genome"){
+    stop("Currently, only genome samples are available for this function. Please select a valid seq type (i.e genome). Compatibility for capture samples will be added soon...")
+  }
+  
+  #get wildcards from this_seq_type (lazy)
+  seq_type = this_seq_type
```
Kdreval, line 1772: I would think this might not be needed here to keep the function simple, and `this_seq_type` can be used as is

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
     chromosome = split_chunks[1]
     startend = unlist(strsplit(split_chunks[2], "-"))
     qstart = startend[1]
     qend = startend[2]
   }
-  if(grepl("chr", chromosome)){
+  
+  #deal with chr prefixes for region, based on selected genome projection.
+  if(projection == "grch37"){
+    if(grepl("chr", chromosome)){
```
Kdreval, line 1786: I think this ifelse is not needed. The `gsub` will not complain when the "chr" is not in the string. Instead, the user here will get warning that only first element is used since the `chromosome` is a vector. I think this part can be reworked to something like
```
always gsub chr substring
if(projection == "hg38"){
    add the chr prefix back using paste0
}
```
This way, there are no ifelse checks so the code is simpler, prefix is properly handled, and user will get no warnings

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
+    if(!grepl("chr", chromosome)){
+      chromosome = paste0("chr", chromosome)
+    }
   }
-
-  #chr prefix the query chromosome to match how it's stored in the table.
-  #This isn't yet standardized in the db so it's just a workaround "for now".
+  
+  #sanity check type of qstart en qend
+  if(!is.numeric(qstart)){
```
Kdreval, line 1796: The ifelse checks can be resource consuming on large data frames with lots of rows. [Here is a different intention ](https://stackoverflow.com/a/41585689/16737475) but it shows that the checks are very slow. Instead, nothing bad will happen if we just always mutate the qstart and qend to numeric regardless of what they initially are.

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
+  
   if(from_flatfile){
-    base_dir = config::get(config="default")$project_base
-
-    unmatched_path = config::get()$results_directories$controlfreec
-
-    #separated by which genome the sample was aligned to
-    unmatched_hg38_path = paste0(unmatched_path, "from--genome--hg38/")
-    unmatched_grch37_path = paste0(unmatched_path, "from--genome--grch37/")
+    cnv_flatfile_template = config::get("results_flatfiles")$cnv_combined$icgc_dart
```
Kdreval, line 1805: We might want to check the user permission on this file here if the user is outside the morinlab, then default to gambl and give a warning. This way, it will not error out for other GSC users who do not have ICGC permissions

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
+    }
+    
+    #check for missingness.
+    if(!file.exists(full_cnv_path)){
+      print(paste("missing: ", full_cnv_path))
+      message("Cannot find file locally. If working remotely, perhaps you forgot to load your config (see below) or sync your files?")
+      message('Sys.setenv(R_CONFIG_ACTIVE = "remote")')
+    }
+        
+    all_segs = all_segs = suppressMessages(read_tsv(full_cnv_path)) %>%
```
Kdreval, line 1825: This here has double assignment, maybe result of the merge? Needs to be resolved - not sure how it will behave

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
-        dplyr::mutate(method = "battenberg")
-
-      # get controlfreec segments for samples with missing battenberg results like unpaired
-      all_segs_unmatched = dplyr::tbl(con, table_name_unmatched) %>%
-        dplyr::filter((chrom == chromosome & start <= qstart & end >= qend) | (chrom == chromosome & start >= qstart & end <= qend)) %>%
-        as.data.frame() %>%
-        dplyr::filter(! ID %in% all_segs_matched$ID)  %>%
-        dplyr::mutate(method = "controlfreec")
-
+    all_segs_matched = all_segs_matched = dplyr::tbl(con, table_name) %>%
```
Kdreval, line 1835: same double assignment here

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
+      dplyr::filter((chrom == chromosome & start <= qstart & end >= qend) | (chrom == chromosome & start >= qstart & end <= qend)) %>%
+      as.data.frame() %>%
+      dplyr::mutate(method = "battenberg")
+      
+    all_segs_unmatched = all_segs_unmatched = dplyr::tbl(con, table_name_unmatched) %>%
```
Kdreval, line 1840: and here

R/database.R
```diff
@@ -1715,98 +1715,154 @@ get_sample_cn_segments = function(this_sample_id,
...
-                           with_chr_prefix = FALSE,
+                           projection = "grch37",
+                           this_seq_type = "genome",
+                           remove_chr_prefix = FALSE,
                            streamlined = FALSE,
-                           from_flatfile = FALSE){
+                           from_flatfile = TRUE){
+  
+  #checks
   remote_session = check_remote_configuration()
```
Kdreval, line 1764: Should this be `auto_connect=TRUE` here?

R/viz.R
```diff
@@ -1608,6 +1613,9 @@ prettyOncoplot = function(maftools_obj,
   if (! is.null(custom_colours)){
     colours = custom_colours
   }
+
+  colours[["Silent"]] = "#A020F0"
```
Kdreval, line 1617: Maybe this should be added elsewhere to the function that handles all colors for mutations? That way, we store all color assignments in one place

R/viz.R
```diff
@@ -2041,7 +2049,9 @@ ashm_multi_rainbow_plot = function(regions_bed,
   p = muts_anno %>%
         ggplot() +
         geom_point(aes(x = start, y = sample_id, colour = classification), alpha = 0.4, size = 0.6) +
-        theme(axis.text.y = element_blank()) +
+        labs(title = "", subtitle = "", x = "", y = "Sample") +
+        theme_cowplot() +
```
Kdreval, line 2053: can we switch to `theme_Morons()` here?

R/viz.R
```diff
@@ -2041,7 +2049,9 @@ ashm_multi_rainbow_plot = function(regions_bed,
   p = muts_anno %>%
         ggplot() +
         geom_point(aes(x = start, y = sample_id, colour = classification), alpha = 0.4, size = 0.6) +
-        theme(axis.text.y = element_blank()) +
+        labs(title = "", subtitle = "", x = "", y = "Sample") +
```
Kdreval, line 2052: I am more of a fan of specifying the empty ones in the `theme()` and setting to `element_blank()`. The reason is that when we set it to nothing, it is still processed as "invisible string" and the space for that `` string is allocated on the plot, creating unnecessary white space. Instead, when set through the theme, the title is removed completely and the plotting area has minimum of white space

R/viz.R
```diff
@@ -2251,19 +2267,26 @@ ashm_rainbow_plot = function(mutations_maf,
...
-    height = length(unique(meta_arranged$sample_id)) + 8
-    p = p + geom_rect(data = bed, aes(xmin = start, xmax = end, ymin = 0, ymax = height + 5), alpha = 0.1) +
+    height = length(unique(meta_arranged$sample_id)) + 10
+    p = p + geom_rect(data = bed, aes(xmin = start, xmax = end, ymin = 0, ymax = height + 20), alpha = 0.1) +
       geom_text(data = bed, aes(x = midpoint, y = height, label = name), size = 2.5, angle = 90) +
       guides(color = guide_legend(reverse = TRUE, override.aes = list(size = 3)))
   }
+
+  p = p +
+    labs(title = "", subtitle = "", x = "", y = "Sample") +
```
Kdreval, line 2277: same comments as above for the multirainbow plot

R/viz.R
```diff
@@ -3437,7 +3462,7 @@ fancy_v_count = function(this_sample,
   }
 
   #add chr prefix if missing
-  if(!str_detect(maf$Chromosome, "chr")){
+  if(!str_detect(maf$Chromosome, "chr")[1]){
```
Kdreval, line 3465: Same logic can be applied here for the chr handling

R/viz.R
```diff
@@ -3694,7 +3719,7 @@ fancy_v_sizedis = function(this_sample,
...
```
Kdreval, line 3722: and here :smiley:

R/database.R
rdmorin: Yeah I was confused about that too. with_chr_prefix seemed reasonable to me because it means the user wants the result to have chr prefix there. Prepend/remove implies an action will be done. The "with" naming implies it will be done, when necessary, if requested (if that makes any sense)

R/database.R
rdmorin: Creating a variable with the name that matches the wildcard is necessary when we use glue. Don't change this unless we're certain glue isn't being used later

R/database.R
rdmorin: The else part is needed because it's prepending chr if (and only if ) it's missing though

R/database.R
rdmorin: good point

R/database.R
mattssca: Those are good points, I will revert the name change here. However, I do still think the old name is somewhat misleading. This is because this parameter only has one condition in the code, if set to FALSE (default) chr prefixes will be removed. At least to me, this implies that if set to TRUE, chr prefixes should be available. But that's not how the function operates. I can easily update this so that if set to TRUE, "chr" will be added (if not already there, i.e if returned segments are in respect to hg38). Does that make sense?

R/database.R
mattssca: Indeed, I'll go over the examples and ensure parameters called with default values are not shown.

R/database.R
mattssca: Precisely, that is indeed why that variable was created. I will not implement the edit suggested by Kostia.

R/viz.R
mattssca: Indeed we can!

R/database.R
mattssca: Oh, yes. Must've been a merge result. I'll go over all of these. Good catch!

R/viz.R
mattssca: Yes, I follow and agree that this is probably the correct way of doing this. Thanks for the comment.

R/database.R
rdmorin: Yes that makes sense. Perhaps this functionality just wasn't added when this function was updated to support hg38?

R/database.R
Kdreval: Right, I meant that it is only checking the first occurrence. Here it is str_detect but grepl I think works the same way. So If there will be a mix of prefixed and non-prefixed chromosome names, the way it is currently implemented might not always deal with this properly
```
Warning message:
In if (!str_detect(genes$chromosome, "chr")) { :
  the condition has length > 1 and only the first element will be used
```

Kdreval (approved):
Thanks Adam! I think all the comments were addressed! I have also tested in on the GSC and I think it is ready to go!

PR #146, opened 2022-12-17
Resolving issues, bugs, and making reproducible installation

Kdreval:
This PR fixes issues #155 and #154. The autoinstallation has been tested both on Mac and Linux-based systems and the set of command that worked is:
```
conda create -n gamblr-dependencies r-base=4.1.3 r-devtools r-biocmanager cmake r-rmysql r-git2r mysqlclient r-tidyverse r-nloptr r-xml -c conda-forge -c anaconda

conda activate gamblr-dependencies

R --vanilla

devtools::install_github("morinlab/GAMBLR@kdreval-dev")
```

In addition to this, there are updates to the GAMBLR color palette because one of the palettes was having duplicated colors for 2 different genetic subgroups. The `prettyForestplot` has also now functionality to report the number of mutated and unmutated patients in the output of the Fisher test results. There are also some other minor bug fixes in `FtestCNV` and `adjust_ploidy` functions.

All documentation was updated accordingly and the most recent master was merged to ensure no conflicts.


Kdreval:
This is ready for review!

R/utilities.R
```diff
@@ -4406,9 +4362,9 @@ classify_fl <- function(
 
 }
 
-#' Classify DLBCLs according to genetic subgroups of Chapuy et al.
+#' Classify DLBCLs according to genetic subgroups.
 #'
-#' Use the feature weights from NMF model to assemble the binary matrix and classify DLBCL tumors based on C0-C5 system of Chapuy et al
+#' Use the feature weights from NMF model to assemble the binary matrix and classify DLBCL tumors
```
rdmorin, line 4367: Can you make this more detailed? What NMF model are these coming from? Is NMF really a model or do you mean from the NMF solution?

rdmorin: I see that this was made less descriptive when you overloaded the function. I think you need to clarify what it does for each classifier now that this is generalized

Kdreval: I have added more descriptions here. Also introduced an option for N1 group in Lacy classifier

Kdreval: This is ready for re-review now!

R/utilities.R
```diff
@@ -4418,25 +4371,29 @@ classify_fl <- function(
 #' @param projection The projection of the samples. Only used to retrerive data through GAMBLR when it is not provided. Defaults to grch37.
 #' @param output The output to be returned after prediction is done. Can be one of predictoins, matrix, or both. Defaults to predictions.
 #' @param adjust_ploidy Whether to perform ploidy adjustment for the CNV data. Defaults to TRUE (recommended).
+#' @param include_N1 When running Lacy classifieer, indicate whether to set samples with NOTCH1 mutations to N1 group as described in Runge et al (2021). Defaults to FALSE.
```
rdmorin, line 4374: Clarify that it's NOTCH1 truncating (I guess frameshift and nonsense?)

R/utilities.R
```diff
@@ -4406,9 +4355,13 @@ classify_fl <- function(
...
 }
 
-#' Classify DLBCLs according to genetic subgroups of Chapuy et al.
+#' Classify DLBCLs according to genetic subgroups.
 #'
-#' Use the feature weights from NMF model to assemble the binary matrix and classify DLBCL tumors based on C0-C5 system of Chapuy et al
+#' Using the user-provided or GAMBLR-retrieved data, this function will assemble the matrix according to the approach of
+#' Chapuy et al (2018) or Lacy et al (2020) classifiers. Since neither of this classifiers is publicly released, we have implemented a solution
+#' that closely (> 92% accuracy) recapitulates each of these systems. For the classifier of Chapuy et al, the constructed matrix will be
+#' used to calculate class probability using the feature weights from bundled NMF model. For the Lacy et al classifier, the matrix will be
```
rdmorin, line 4363: Clarify what a bundled NMF model is. I thought NMF was clustering and your models were all RF

R/utilities.R
```diff
@@ -4517,23 +4474,78 @@ classify_dlbcl_chapuy <- function(
...
+    # Now use this data to classify the samples according to one of the systems
+    if(method=="chapuy"){
+
+      # SVs are needed only for Chapuy predictions
+      # If no SV data is provided, get the SVs from GAMBL
+      if(missing(sv_data)){
+          message("No SV data is provided.")
+          message("Will retreive SVs available through GAMBL.")
+
+          sv_data <- get_combined_sv(
```
rdmorin, line 4487: This is using a function that doesn't have the complete SV data (last updated in October). We should address this

R/utilities.R
```diff
@@ -4517,23 +4474,78 @@ classify_dlbcl_chapuy <- function(
...
+      predictions <- classify_dlbcl_lacy(
+          these_samples_metadata,
+          maf_data,
+          seg_data,
+          projection = projection,
+          output = output,
+          include_N1 = include_N1
+      )
+    }else{
+      stop("You requested an unvalid method. Please choole one of chapuy or lacy")
```
rdmorin, line 4519: Choose

R/utilities.R
```diff
@@ -4418,25 +4371,29 @@ classify_fl <- function(
...
 #' @return data frame with classification, binary matrix used in classification, or both
 #' @export
 #' @import data.table circlize tidyverse
 #'
 #' @examples
 #' test_meta <- get_gambl_metadata(case_set = "DLBCL-unembargoed")
-#' predictions <- classify_dlbcl_chapuy(these_samples_metadata = test_meta)
-#' matrix_and_predictions <- classify_dlbcl_chapuy(these_samples_metadata = test_meta, output = "both")
+#' predictions_chapuy <- classify_dlbcl(these_samples_metadata = test_meta)
+#' predictions_lacy <- classify_dlbcl(these_samples_metadata = test_meta, method = "lacy")
```
rdmorin, line 4383: How about calling it "hmrn" or "HMRN" as a fourth "model" that is really just using lacy + N1?

R/utilities.R
```diff
@@ -4517,23 +4474,78 @@ classify_dlbcl_chapuy <- function(
...
+#' Classify DLBCLs according to genetic subgroups of Chapuy et al.
+#'
+#' Use the feature weights from NMF model to assemble the binary matrix and classify DLBCL tumors based on C0-C5 system of Chapuy et al
+#'
+#' @param these_samples_metadata The metadata data frame that contains sample_id column with ids for the samples to be classified.
+#' @param maf_data The MAF data frame to be used for matrix assembling. At least must contain the first 45 columns of standard MAF format.
+#' @param seg_data The SEG data frame to be used for matrix assembling. Must be of standard SEG formatting, for example, as returned by get_sample_cn_segments.
+#' @param sv_data The SV data frame to be used for matrix assembling. Must be of standard BEDPE formatting, for example, as returned by get_combined_sv.
+#' @param projection The projection of the samples. Only used to retrerive data through GAMBLR when it is not provided. Defaults to grch37.
+#' @param output The output to be returned after prediction is done. Can be one of predictoins, matrix, or both. Defaults to predictions.
```
rdmorin, line 4537: fix predictoins

rdmorin, line 4537: Instead of "both" as an option what about just returning both by default if the user doesn't specify "output" explicitly?

R/utilities.R
```diff
@@ -4275,41 +4277,6 @@ classify_fl <- function(
...
-        )
-        message(
-            "They will be set to 0, which may affect model performance."
-        )
-        ssm_matrix[,ssm_missing] <- 0
-    }
-
     # Generate binary matrix for ashm
     ashm_matrix <- get_ashm_count_matrix(
         ashm_features_bed,
```
rdmorin, line 4282: What is being used for this bed file? Seems like a specific file is needed but where does it come from?

rdmorin (changes requested):
A few things to address or consider addressing

R/utilities.R
Kdreval: It is defined a bit more up, at the line 4184:
```
    ashm_features_bed <- grch37_ashm_regions %>%
        dplyr::mutate(
            name = gsub("-", "_", grch37_ashm_regions$name)
        ) %>%
        dplyr::filter(name %in% ashm_features | name %in% c("PAX5_TSS_1", "SGK1_TSS_1"))
```

R/utilities.R
Kdreval: Will be updated in the next commit

R/utilities.R
Kdreval: This is what I was trying to figure out yesterday too. Lacy/Runge considered only frameshift, and this is what I implemented here too. The truncating status is now clarified and I will push this update

R/utilities.R
Kdreval: Great idea! Changed to this and implemented this update

R/utilities.R
Kdreval: User can provide their own SV data in bed format, this is just resorting to use what is available if nothing else is given.

R/utilities.R
Kdreval: This is now fixed

R/utilities.R
Kdreval: Fixed!

Kdreval: I have changed this to both so the default output now is matrix and the predictions

R/utilities.R
Kdreval: I took a look how this will compare vs when get_manta_sv is used. There is a bit of a switch. I think I will switch this to manta since this will probably be more up to date going forward and the user can supply the output of combined sv if they wish

Kdreval: This is now switched to the manta svs :heavy_check_mark:

Kdreval:
Also confirmed the PR approved yesterday has no merge conflicts with this one

Kdreval:
This PR now also fixes the bug of `randomForest` dependency auto-installation reported in the issue #154

Kdreval:
In this PR now I have moved out the functionality of classifications and predictions to it's own designated R package [GAMLR.predict](https://github.com/morinlab/GAMBLR.predict). It is ready for review and merge if all looks good. Thanks!

R/viz.R
```diff
@@ -2727,10 +2728,63 @@ prettyForestPlot = function(maf,
     dplyr::mutate(test = map(table, fisher.test), tidy = map(test, broom::tidy)) %>%
     unnest(tidy) %>%
     dplyr::mutate(q.value = p.adjust(p.value, "BH")) %>%
-    dplyr::select(-c(table, test, method, alternative)) %>%
+    dplyr::select(-c(test, method, alternative)) %>%
     dplyr::filter(q.value <= max_q) %>%
     dplyr::mutate(gene = fct_reorder(gene, estimate))
 
+  flatten_table <- function(Row){
```
Kdreval, line 2735: This functionality is added so the `fisher` result of calling  this function now also reports 4 additional columns: number of mutated and unmutated samples for each comparison group

NAMESPACE
```diff
@@ -140,10 +136,7 @@ import(maftools)
 import(metaviz)
 import(parallel)
 import(plotly)
-import(randomForest)
-import(reshape2)
 import(rtracklayer)
-import(stats)
```
mattssca, line 146: From the documentation overhaul that I am working on, I think this package is still being used by at least one existing GAMBLR function (`FtestCNV`). I could be wrong, what are your thoughts?

NAMESPACE
```diff
@@ -140,10 +136,7 @@ import(maftools)
 import(metaviz)
 import(parallel)
 import(plotly)
-import(randomForest)
```
mattssca, line 143: I guess this fix solves issue #160. If that's the case, I can revert the fix I added to my branch (i.e [this](https://github.com/morinlab/GAMBLR/pull/164/files#diff-2ee57637b77c38eb625492f80e8c6d88b08103cce85eb89198a2e7f2f3b1832fR2154) and [this](https://github.com/morinlab/GAMBLR/pull/164/files#diff-2ee57637b77c38eb625492f80e8c6d88b08103cce85eb89198a2e7f2f3b1832fR2396))

mattssca (commented):
I think this looks good. Just a few minor comments. Thanks Kostia!

FYI, I have recently pushed an updated version of the [utilities.R](https://github.com/morinlab/GAMBLR/pull/164/commits/665cbee39e17031988842dd1674a18db7365dffc) script to my branch. This update includes a complete documentation overhaul. So if you want to use the function documentation for the removed functions in the new R package, you could grab them from there, if you're not interested in writing them yourself. 

Thanks again!

NAMESPACE
Kdreval: Please keep your fixes. It is a good practice to refer to the function through the package. I did not mean to "override" your fixes - they will still be needed to ensure reproducibility.

NAMESPACE
Kdreval: Thanks for noticing, stats should be dropped from that function too. I'll push the fix!

NAMESPACE
mattssca: For sure!

NAMESPACE
Kdreval: this is now done ✅

mattssca:
In addition, I also think `prettyForestPlot` is using `stats` functions (p.adjust) and `prettyOncoplot` might also be using the same package (calling the `setNames` function). Thoughts?

Kdreval:
When doing this update I found the `stats` package is one of the r default packages. It is always loaded on startup and is always available without explicitly loading it. When it was added as a dependency, it only created issues with name conflicts etc. Now having this removed makes the GAMBLR easier to load and instead of ~40 warnings on startup, there are only ~ 20. I would think we do not need to revert it back to being explicitly mentioned in NAMESPACE and DESCRIPTION.

mattssca:
Sounds good Kostia, thanks for looking into it!

mattssca (approved):

Kdreval:
@rdmorin the merge is blocked even after approval because earlier some changes were requested. Can you please look into this?

rdmorin (approved):

rdmorin:
Should work now

Kdreval:
thanks!!

Issue #145, opened 2022-12-09
Is genes an optional or required argument for prettyForestPlot + other prettyForestPlot issues


rdmorin:
The documentation says that `genes` is optional but when I run this function without it I get an error. Can the function be fixed to work without this option specified and/or the docs be updated to make this required?

```
prettyForestPlot(maf=bl_coding,metadata=bl_meth_meta,comparison_column="cluster",comparison_values=c("cluster1","cluster2"))
 Error in h(simpleError(msg, call)) : 
error in evaluating the argument 'table' in selecting a method for function '%in%': argument "genes" is missing, with no default
```

mattssca:
I think it's best to update the documentation for this function so that it is clear that `genes` is a required parameter. 

Specified genes in this parameter are required to filter the incoming maf and this is what the error you are seeing relates to. 

In addition, there is a check at the beginning of this function that is meant to stop the function if no genes are provided. However, this currently does not work as intended. Will be fixed in a PR dedicated to resolving the issues raised for this function.

mattssca:
There's another issue related to this plot:
```
prettyForestPlot(maf=bl_coding,metadata=bl_meth_meta,comparison_column="cluster",comparison_values=c("cluster1","cluster2"),genes=dlbcl_genes,max_q = 0.1)
FONT: 20 POINT: Inf 0
Error in `check_aesthetics()`:
! Aesthetics must be either length 1 or the same as the data (1): ymin, ymax, x and y
Run `rlang::last_error()` to see where the error occurred.
```

rdmorin:
Wouldn't it be better to have it work as advertised though? Some folks may want to run the test on all genes rather than have to subset them upfront

mattssca:
Of course, that sounds reasonable. I can add this functionality, no problem.

rdmorin:
Resolved?

mattssca:
Resolved!

PR #148, opened 2022-12-22
GAMBLR Improvements

mattssca:
# **Updates included in this PR**
## Major changes:
1. `sv_to_custom_track` up until now, does not work with the parameter `is_annotated = FALSE`. Possible due to an old auto-merge or remnants of old code. Maybe this function is not called as frequently, allowing the bug to unknowingly exist. The function was resolved and new examples were added (for both Boolean cases with the `is_annotated` parameter).

2. `gene_to_region` and `region_to_gene` have been reverted to use locally stored gene coordinates and not rely on biomartR (unable to get the desired ensemble version (86) of gene annotations). Previously no hg38 gene coordinates were available in GAMBLR, restricting the just mentioned functions to only work with grch37. Gene annotations with respect to hg38 have been added to `/GAMBLR/data/hg38_gene_coordiantes.Rda`. Script on how these coordinates were retrieved and cleaned up has been added to `/data-raw/DATASET.R` for reproducibility purposes. 

3. Updates to `prettyForest` plot. I was unsuccessful in reproducing the [error](https://morinlabsfu.slack.com/archives/C0224H120CU/p1670607082425289) Ryan [reported](https://github.com/morinlab/GAMBLR/issues/145) last week (even with his exact datasets). So not sure what is going on here. Maybe Rayan can have another go at it and see if the error persists. This function has also been updated to now work without a gene list provided. If this parameter is not used, the function will extract all mutated genes from the incoming maf and with another optional parameter, `n_mutations`, the user can now easily subset this gene list to any given mutation frequency, e.g set to 20 to only keep genes that are mutated in 20 (or more) samples. Allowing the user to reduce the number of test runs, giving a better chance of a significant result. Attached is the plot I generated using Ryan’s datasets, as well as the source code.

```
bl_meth_meta = read_tsv(file = "../../bl_meth_meta.tsv")
bl_coding = read_tsv(file = "../../bl_coding.tsv")

dlbcl_genes = lymphoma_genes %>%
  dplyr::filter(DLBCL == TRUE) %>%
  pull(Gene)

prettyForestPlot(maf = bl_coding,
                 metadata = bl_meth_meta,
                 comparison_column = "cluster",
                 comparison_values = c("cluster1", "cluster2"),
                 genes = dlbcl_genes,
                 max_q = 0.1)
```

<p align="center">
  animated
</p>

4. New functions, `get_manta_sv_by_sample` and `get_manta_sv_by_samples` (wraps the first out of the two functions). For the design of these functions, I picked up Ryan's active PR (#143) and expanded upon this code. Relevant VCF information (based on returned columns for `get_manta_sv`) has also been added, as well as multiple checks, examples, documentation, etc. I have tested these functions with a variety of parameter combinations and the returns are satisfactory. For the reviewer's convenience, below is a sample output of calling `get_manta_sv_by_sample` on one sample, testing is obviously still encouraged.

```
> my_bedpe_df = get_manta_sv_by_sample(this_sample_id = "00-14595_tumorA")
[1] "Reading from: /projects/nhl_meta_analysis_scratch/gambl/results_local/gambl/manta_current/99-outputs/bedpe/genome--grch37/somaticSV/00-14595_tumorA--00-14595_normal--matched.somaticSV.bedpe"                    
> head(my_bedpe_df)
  CHROM_A  START_A    END_A CHROM_B  START_B    END_B NAME SOMATIC_SCORE STRAND_A STRAND_B TYPE FILTER VAF_tumour VAF_normal DP_tumour DP_normal tumour_sample_id normal_sample_id pair_status
1       1  1556541  1556547       1  1556664  1556670    .            40        -        -  BND   PASS      0.145          0        55        73  00-14595_tumorA  00-14595_normal     matched
2       1  6012725  6012732       1  6012825  6012832    .            48        +        +  BND   PASS      0.156          0        32        85  00-14595_tumorA  00-14595_normal     matched
3       1  8464072  8464090       1  8464293  8464311    .            40        +        +  BND   PASS      0.112          0       215        84  00-14595_tumorA  00-14595_normal     matched
4       1 10084099 10084251       1 10084266 10084411    .            48        -        -  BND   PASS      0.186          0        43        50  00-14595_tumorA  00-14595_normal     matched
5       1 10526162 10526172       1 10526290 10526300    .            40        -        -  BND   PASS      0.157          0       121        92  00-14595_tumorA  00-14595_normal     matched
6       1 15878430 15878436       1 15878608 15878614    .            40        +        +  BND   PASS      0.119          0       126       101  00-14595_tumorA  00-14595_normal     matched
```

## Minor changes:
5. `fancy_v_chrcount` now internally calls `collate_results` and not `collate_qc_results`.

6. Parameter description updates for `get_combined_sv` to accurately represent the functionality of the projection parameter.

7. Fixin typos and missing parameter descriptions to `get_sample_cn_segments`.

8. Updating parameter descriptions for `get_manta_sv`.

# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


R/database.R
```diff
@@ -1594,8 +1594,10 @@ get_cn_states = function(regions_list,
 #' Get all segments for a single (or multiple) sample_id(s).
 #'
```
rdmorin, line 1595: The parameters for this function should be similar (or the same) as other functions that work this way. Right now, a user must supply 
this_sample_id="some_id" or 
multiple_samples = TRUE, sample_list = c("some_sample", "another_sample") 

Why? 
Wouldn't it make more sense to ask the user to supply either this_sample_id or  these_samples_metadata and let the function figure out if they want data from multiple samples based on how many samples are in these_samples_metadata?

R/database.R
```diff
@@ -2628,3 +2630,256 @@ get_gene_expression = function(metadata,
...
+                                   min_score = 40,
+                                   pass = TRUE,
+                                   with_chr_prefix = FALSE,
+                                   from_flatfile = TRUE,
+                                   projection = "grch37"){
+
+  #check remote configuration
+  remote_session = check_remote_configuration(auto_connect = TRUE)
+
+  #check selected projection
```
rdmorin, line 2670: Needs to support both

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+#'
+#' This is a convenience wrapper function for get_manta_sv_by_sample. See that function for more information.
+#'
+#' @param these_samples_metadata The only required parameter is a metadata table (data frame) that must contain a row for each sample you want the data from. The additional columns the data frame needs to contain, besides sample_id, are: unix_group, genome_build, seq_type, pairing_status
+#' @param min_vaf The minimum tumour VAF for a SV to be returned.
+#' @param min_score The lowest Manta somatic score for a SV to be returned.
+#' @param pass If set to TRUE, include SVs that are annotated with PASS in FILTER column. Default is TRUE.
+#' @param with_chr_prefix Prepend all chromosome names with chr (required by some downstream analyses).
+#' @param from_flatfile Set to TRUE by default.
+#' @param projection The projection genome build.
```
rdmorin, line 2645: I am having a hard time following the whole all_sv_hg83/all_sv_grch37 setup. Seems like you're making the user do the work and supply what's available. Why doesn't the function work this out for the user?


There's also a typo in your documentation (83 instead of 38).

rdmorin: I am guessing you did this so this function doesn't need to call get_manta_sv multiple times. Is that right?

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+                                   with_chr_prefix = FALSE,
+                                   from_flatfile = TRUE,
+                                   projection = "grch37",
+                                   all_sv_grch37,
+                                   all_sv_hg38){
+
+  #check remote configuration
+  remote_session = check_remote_configuration(auto_connect = TRUE)
+
+  #load merged files if not provided
```
rdmorin, line 2674: I see no reason to make these parameters. Why wouldn't this function just load these files automatically? It only needs to happen once, right? It's confusing to have these new parameters and unnecessary confusion if they are rarely (or never) going to be used

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+  if(missing(all_sv_hg38)){
+    all_sv_hg38 = get_manta_sv(projection = "hg38")
+  }
+
+  #create an empty list.
+  all_bedpe = list()
+
+  #get sample IDs from provided metadata.
+  samples = pull(these_samples_metadata, sample_id)
+
```
rdmorin, line 2688: The line below only needs to be run for samples that are not represented in the all_sv_PROJECTION for whichever build was requested. Seems like the easiest thing to do here is to make a copy of the metadata and drop all the lines that match samples we have SV results from. Then call the get_manta_sv_by_sample for what's left

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+                                  from_flatfile = TRUE,
+                                  projection = "grch37",
+                                  all_sv_grch37,
+                                  all_sv_hg38){
+
+  #check remote configuration
+  remote_session = check_remote_configuration(auto_connect = TRUE)
+
+  #retrieve sample ID from metadata, or get metadata for provided sample ID
+  #get sample ID from metadata and check if metadata contains more than 1 row
```
rdmorin, line 2758: This is a very convoluted approach that will cause this function to be called a bunch of times unnecessarily. Extra overhead that can be avoided. See my comments on the function that calls this one.

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+  }else{
+    normal_sample_id = config::get("unmatched_normal_ids")[[unix_group]][[seq_type]][[genome_build]]
+  }
+
+  if(from_flatfile){
+    if(!remote_session){
+      path_template = config::get("results_flatfiles")$sv_manta$template
+      path_template = paste0(config::get("project_base"), path_template)
+      bedpe_path = glue::glue(path_template)
+
```
rdmorin, line 2802: The logic here seems to be backwards. This will use the individual flatfiles as long as they exist even though most of those results are already in the merge. It should be the other way around: default to using what's in the merge and resort to the flatfiles if the sample is not in the merge.

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+              dplyr::filter(normal_sample_id == this_sample_id) #look for this_sample_id in the normal_sample_id column.
+            if(!nrow(bedpe_dat==0)){ #if sample not available, look for it in the tumour_sample_id.
+              bedpe_dat = all_sv_hg38 %>%
+                dplyr::filter(tumour_sample_id == this_sample_id)
+            }
+          }
+        }
+
+        #if can't find the sample in the existing merge, look for individual files (in the opposite projection) and lift over to the desired projection.
+        if(!nrow(bedpe_dat==0)){
```
rdmorin, line 2827: This code is assuming that if a file is missing then the genome build must be opposite to the projection. The actual genome_build is in the metadata. What if a file is missing for some other reason?

R/database.R
```diff
@@ -2628,3 +2630,385 @@ get_gene_expression = function(metadata,
...
+            bedpe_path = glue::glue(path_template)
+            genome_build = "grch37"
+            bedpe_dat = liftover_bedpe(bedpe_file = bedpe_path, target_build = "grch37")
+
+          }else if(projection == "hg38"){
+            genome_build = "grch37"
+            bedpe_path = glue::glue(path_template)
+            genome_build = "hg38"
+            bedpe_dat = liftover_bedpe(bedpe_file = bedpe_path, target_build = "hg38")
+          }
```
rdmorin, line 2839: As you implemented it, this function is going to be running liftover many times on the individual files rather than just once, right?

rdmorin (changes requested):
I am confused by the logic you used in implementing this. Many pointless function calls and a lot of small tasks that could be handled in one big task if implemented differently. Let me know if you need further clarification after you read through the comments.

R/database.R
mattssca: Yes, That is why I added this as a parameter. For this function, I have now removed these parameters. The function now loads all manta sv calls, based on the selected projection.

R/database.R
mattssca: For this function, I have removed the two parameters `all_sv_grch37` and `all_sv_hg38` and replaced with a new optional parameter called, `all_sv`. If not provided, the function will get it for you, based on the selected genome projection.

R/database.R
mattssca: So I thought the purpose of this function was to be able to retrieve manta sv calls for multiple samples and that what you are suggesting should be its own, more generic function i.e `get_all_sv`. However,  I agree that the suggested approach above should work well for this, I had thought of something similar. Maybe it could all be done with this function as you're suggesting, say if no `these_samples_metadata` is provided, the function operates on the missing samples (make a copy of the metadata and drop all the lines that match samples we have SV results from).

R/database.R
mattssca: Ok, I am with you. A misunderstanding on my part. This has been addressed.

R/database.R
mattssca: Yes, I added this as a "last resort" type of step, if the sample is not available in the merge, or individual flat-file -> look for the sample (flat-file) in the opposite genome projection and lift it over to the desired projection. I can add another statement notifying the user if this requested file (flat file from the opposite projection) is not available (i.e missing for some other reason).

R/database.R
mattssca: Yes, I couldn't think of any other way to do this. I know ideally, we would want to use liftover_bedpe on all sv calls that would need to lifted (instead of running this on individual files). Maybe it would be all right since this step would only be used in "rare cases" when a sample is not available in the merge, or individual flatfile (in the desired projection)? If not, I would be happy to take suggestions on how this can be achieved.

R/database.R
mattssca: I agree. in This PR, I have not touched this function other than updating the already existing parameter descriptions, but they should be updated as you're suggesting, of course.

R/database.R
```diff
@@ -522,10 +522,10 @@ get_merged_result = function(tool_name,
                              projection = "grch37",
                              seq_type = "genome"){
```
rdmorin, line 524: Is this function actually used by any other functions or should it be deprecated entirely?

R/database.R
```diff
@@ -1248,22 +1248,17 @@ get_manta_sv = function(min_vaf = 0.1,
...
+    region = gsub(",", "", region)
+    #format is chr6:37060224-37151701
+    split_chunks = unlist(strsplit(region, ":"))
+    chromosome = split_chunks[1]
+    startend = unlist(strsplit(split_chunks[2], "-"))
+    qstart = startend[1]
+    qend = startend[2]
+  }
 
   #this table stores chromosomes with un-prefixed names. Convert to prefixed chromosome if necessary
```
rdmorin, line 1263: I'm realizing now why the remote functionality isn't working. The bundled snakefile currently doesn't sync the merged manta file locally. I think this is because I was assuming get_combined_sv would entirely replace get_manta_sv

R/database.R
```diff
@@ -1279,6 +1274,16 @@ get_manta_sv = function(min_vaf = 0.1,
 
     all_sv = suppressMessages(read_tsv(sv_file, col_types = "cnncnncnccccnnnnccc", col_names = cnames))
```
rdmorin, line 1275: To address the issue I raised above, can you just replace this line with a call to get_combined_sv()
e.g.
all_sv = get_combined_sv()

R/database.R
```diff
@@ -2628,3 +2636,258 @@ get_gene_expression = function(metadata,
...
+  #subset calls that does not need a "lift"
+  no_lift_needed = merged_bedpe %>%
+    dplyr::filter(need_lift == FALSE)
+  
+  #combine calls (lifted and not lifted), arrange and sort accordingly, drop temporary column
+  merged_bedpe = rbind(lifted_calls, no_lift_needed) %>%
+    arrange(CHROM_A, CHROM_B, START_A) %>%
+    dplyr::select(-need_lift)
+  
+  #return merged manta SVs.
```
rdmorin, line 2706: Not sure why there are rownames that need to be dropped. Is this because you used rbind instead of bind_rows maybe?

R/database.R
```diff
@@ -2628,3 +2636,258 @@ get_gene_expression = function(metadata,
...
+#' @param pass If set to TRUE, include SVs that are annotated with PASS in FILTER column. Default is TRUE.
+#' @param with_chr_prefix Toggle if chromosome names should be prefixed or not, default is FALSE.
+#' @param projection The projection of returned calls. Default is grch37.
+#'
+#' @return a data frame containing the Manta outputs from this_sample_id in a bedpe-like format with additional columns extracted from the VCF column.
+#' @import tidyverse
+#' @export
+#'
+#' @examples
+#' my_svs = get_manta_sv_by_sample(this_sample_id = "00-14595_tumorD",
```
rdmorin, line 2733: This example is overly complicated, right? The metadata doesn't need to be subset like this since the function handles that for you.

R/database.R
```diff
@@ -2628,3 +2636,258 @@ get_gene_expression = function(metadata,
...
+#'                              these_samples_metadata = get_gambl_metadata() %>% dplyr::filter(sample_id == "00-14595_tumorD"))
+#'
```
rdmorin, line 2735: Why is the default force_lift = TRUE ?

R/database.R
```diff
@@ -2628,3 +2636,258 @@ get_gene_expression = function(metadata,
...
+      suppressMessages(suppressWarnings(dir.create(dirN, recursive = T)))
+      ssh::scp_download(ssh_session, remote_bedpe_path, dirN)
+    }
+  }
+  
+  #read sample flat-file
+  message(paste0("Reading ", this_sample_id, " from: ", bedpe_path))
+  bedpe_dat_raw = suppressMessages(read_tsv(bedpe_path, comment = "##"))
+    
+  #if returned manta calls data frame is empty, return and potentially move over to the next sample (if get_manta_sv_by_samples is called).
```
rdmorin, line 2796: This is a bit of a confusing comment at the end. The behaviour of what happens next is not dependent on this function so I'd just leave it as "return empty data frame" or something like that

R/database.R
```diff
@@ -2628,3 +2636,258 @@ get_gene_expression = function(metadata,
...
+  if(!nrow(bedpe_dat_raw==0)){
+    message(paste0("WARNING! No SV calls found in flat-file for: ", this_sample_id))
+    return()
+  }
+  
+  #if selected projection is different from the genome build (for a particular sample), add information that this sample needs to be lifted (by get_manta_by_samples).
```
rdmorin, line 2802: What happens if you run this on a sample from the icgc_dart unix group?

R/preprocessing_io.R
```diff
@@ -1014,7 +1015,8 @@ liftover_bedpe = function(bedpe_file,
                           target_build = "grch37",
                           col_names,
                           col_types,
-                          standard_bed = FALSE){
+                          standard_bed = FALSE,
+                          print_head = TRUE){
```
rdmorin, line 1019: This is only for debugging, right? If so please set to FALSE by default. Also consider renaming the parameter to verbose to be consistent with other functions with a verbose print-happy debugging-mode

R/utilities.R
```diff
@@ -242,11 +242,11 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
 #' Get regions from genes.
 #'
 #' @param gene_symbol A vector of one or more gene symbols.
-#' @param ensembl_id A vector of one or more Esemble IDs.
+#' @param ensembl_id A vector of one or more Ensemble IDs.
```
rdmorin, line 245: There is no e at the end of Ensembl

R/utilities.R
```diff
@@ -1093,57 +1069,62 @@ review_hotspots = function(annotated_maf,
 #' @export
 #'
 #' @examples
+#' #custom track withj annotations
```
rdmorin, line 1072: withj ?

R/viz.R
```diff
@@ -2608,9 +2608,10 @@ theme_Morons = function(base_size = 14,
 #' Create a forest plot comparing mutation frequencies for a set of genes between two groups.
 #'
 #' @param maf A maf data frame. Minimum required columns are Hugo_Symbol and Tumor_Sample_Barcode.
-#' @param mutmat Optional argument for binary mutation matrix. If not supplied, function will generate this amtrix from the file used in argument "maf".
+#' @param mutmat Optional argument for binary mutation matrix. If not supplied, function will generate this matrix from the file used in argument "maf".
 #' @param metadata Metadata for the comparisons. Minimum required columns are Tumor_Sample_Barcode and the column assigning each case to one of two groups.
-#' @param genes An optional list of genes to restrict your plot to.
+#' @param genes An optional list of genes to restrict your plot to. If no gene-list is supplied, the function will extract all mutated genes from the incoming maf. See n_mutations parameter for more info.
+#' @param n_mutations Optional parameter for when no gene list is provided. This parameter ensures only genes with n mutations are kept in the gene list. Default value is 1, this means all genes in the incoming maf will be plotted.
```
rdmorin, line 2614: How about min_mutations so it's more self-explanatory?

config.yml
```diff
@@ -84,6 +84,8 @@ default:
             icgc_dart: "all_the_things/cnv_master-1.0/merges/{seq_type}--{projection}.seg"
             template: "all_the_things/cnv_master-1.0/merges/{seq_type}--{projection}.seg"
         sv: "icgc_dart/cbioportal-1.0/01-filtered-manta/genome-grch37/all_merged_sv_info.tsv"
+        sv_manta:
```
rdmorin, line 87: I thought this was already in the config. Maybe just in my branch though?

rdmorin (changes requested):
A few questions and one possible bug. There's also the issue of remote functionality but we can address that later as long as the GSC functionality works.

R/database.R
mattssca: Indeed, this function was only called by `get_manta_sv()`. I have now removed it from GAMBLR.

R/database.R
mattssca: Got it, let's revisit this.

R/database.R
mattssca: This function has now been updated to use `get_combined_sv` (replacing, `get_merged_results`).

R/database.R
mattssca: The row names were dropped since my approach for dealing with imprecise variants scrambled all rows when stitching the data back together. It has no effect on the data frame, I thought it looked cleaner not to have row names in this data frame. However, `bind_rows()` solves this without setting row names to NULL. Both `get_manta_by_x` functions have been updated to use `bind_rows` instead, good suggestion!

R/database.R
mattssca: Yeah, that is true. I believe parts of this example are a remnant of a previous draft of this function, thus requiring a more complex execution. But It has been updated now. Thanks!

R/database.R
mattssca: I explained this in more detail in my weekly update. But anyways, it is because, **if** the user would ever want to run this function on its own, they could still get lifted coordinates back without having to change this parameter. For the `get_manta_by_samples`, this function forces this parameter to be `FALSE`, so the lifting takes place in that function instead. If that makes sense?

R/database.R
mattssca: Sounds good to me!

R/database.R
mattssca: We get SV calls returned for this sample;
```
> icgc_dart_sample = get_gambl_metadata() %>%
+   dplyr::filter(unix_group == "icgc_dart") %>%
+   sample_n(1) %>%
+   pull(sample_id)
                                                                                                                                                                                                                                                                 
> my_svs = get_manta_sv_by_sample(this_sample_id = icgc_dart_sample,
+                                 these_samples_metadata = get_gambl_metadata())
Reading SP116645 from: /projects/nhl_meta_analysis_scratch/gambl/results_local/icgc_dart/manta_current/99-outputs/bedpe/genome--hs37d5/somaticSV/SP116645--SP116644--matched.somaticSV.bedpe                                                                     
SP116645 flat-file is not avaialble in the selected projection, running liftover_bedpe...                                                                                                                                                                        
SP116645 successfully lifted to grch37
> head(my_svs, 10)
   CHROM_A   START_A     END_A CHROM_B   START_B     END_B NAME SOMATIC_SCORE STRAND_A STRAND_B TYPE FILTER VAF_tumour VAF_normal DP_tumour DP_normal tumour_sample_id normal_sample_id pair_status
1        1  12975734  12975737       1  12975837  12975837    .            45        +        -  DEL   PASS      0.250      0.000        52        95         SP116645         SP116644     matched
2       10  90731254  90731254      10  90731688  90731688    .            56        +        -  DEL   PASS      0.130      0.000        69        94         SP116645         SP116644     matched
3       13  72927411  72927418      13  72927489  72927489    .            62        +        -  DEL   PASS      0.395      0.056        38        89         SP116645         SP116644     matched
4       14  74278653  74278653      14  74451363  74451363    .            68        +        -  DUP   PASS      0.104      0.000       134       140         SP116645         SP116644     matched
5       14  84711735  84711737      14  84715223  84715225    .            73        +        -  DEL   PASS      0.205      0.000        88       111         SP116645         SP116644     matched
6       14 106667431 106667433      14 106681734 106681736    .            82        +        -  DEL   PASS      0.193      0.000        83       123         SP116645         SP116644     matched
7       14 106695732 106695732      14 106696312 106696312    .            59        -        -  BND   PASS      0.327      0.000        49        68         SP116645         SP116644     matched
8       14 106696782 106696784      14 106780314 106780316    .            75        +        -  DEL   PASS      0.265      0.000        34        77         SP116645         SP116644     matched
9       14 106781774 106781774      14 106783374 106783374    .            69        +        -  DEL   PASS      0.232      0.000        56        88         SP116645         SP116644     matched
10      14 106783723 106783723      14 106783902 106783902    .            61        +        -  DEL   PASS      0.246      0.000        69        88         SP116645         SP116644     matched
```

R/preprocessing_io.R
mattssca: So this function always prints the header of the lifted bedpe, this really clobbered the terminal when running `get_manta_by_samples` on a large cohort of samples (i.e all the missing samples). That is why I turned the `print(head)` statement into a parameter that is TRUE per default for the liftover function (i.e this function still works the same way as previously but allows us to silence this printing when we're performing liftover on multiple samples). I have changed the parameter name to be "verbose", but I think it should be TRUE per default for this function, to honour the way this function previously worked. Do you think this makes sense?

R/utilities.R
mattssca: I agree, I changed a typo into another typo :)

R/utilities.R
mattssca: Got it, good catch :)

R/viz.R
mattssca: I like it, thanks!

config.yml
mattssca: Yeah, I did not have it in my branch, that is for sure.

mattssca:
Review comments have been addressed and merge conflicts resolved. ready for rereview and merge if no outstanding issues remain.

R/database.R
rdmorin: Is this resolved?

R/database.R
rdmorin: Please mark any of these comments as resolved if they are no longer relevant/dealt with

R/database.R
rdmorin: is this resolved?

R/database.R
rdmorin: is this resolved?

R/database.R
rdmorin: is this resolved?

R/database.R
rdmorin: I think that this may lead to liftOver being run unnecessarily though. How about setting it to FALSE by default and warning the user that if they set it to TRUE they should be sure that the file they need isn't already available from the merge (and suggest they try get_combined_sv)

R/database.R
rdmorin: is this resolved?

R/database.R
rdmorin: So it runs liftOver from hs37d5 to grch37?

R/preprocessing_io.R
rdmorin: No. verbosity should never occur by default. The old print(head) line was probably something that was there for debugging.

rdmorin:
Please set all comment threads that are resolved to "resolved" so I can focus on unresolved ones. The liftover of ICGC data is a bug that needs to be addressed, I think.

R/database.R
mattssca: Ok, I like this. Will add this in a new commit.

R/preprocessing_io.R
mattssca: Ok, I will address this then

R/database.R
mattssca: Yes, it does.

mattssca:
All conversations have been resolved, except for the "liftover of ICGC data" comment.

mattssca:
This function has now been updated to accurately deal with different genome builds (other than grch37 and hg38). 

The function now looks for any discrepancies between genome_build and selected projection for a particular sample, if genome build contains either 37 or 19 in its string (e.g grch37, hg19, hs37d5), this is all equivalent to grch37 (projection). Meaning that if the genom_build for the selected sample is either grch37, hg19, hs37d5 and the selected projection is grch37 or hg19, no lift-over will take place. In the same example, if the selected projection is hg38 or grch38 the variant calls will be lifted to the selected projection. 

Similarly, if the genome_build has 38 in its string and the selected projection is grch37 or hg19, liftover will take place and if the projection (in the same example) is hg38 or grch38 no lift-over will take place. 

I have also removed the parameter with_chr_prefix and automated this step based on the selected projection.

R/database.R
mattssca: When testing, I discovered that the output from `get_combined_sv()`, does not match the output from the previous function (`get_merged_results()`). The columns are quite different, meaning that this code from `get_manta_sv` fails; `all_sv = rbind(all_sv, missing_sv)`. 

The question is, should we revert and still `call get_merged_results` in `get_manta_sv`, or should the return for `get_manta_by_sample` be modified so that it matches what gets returned with `get_combined_sv`?

Kdreval: I would go with option 2!

mattssca:
I've pushed updates to the data wrangling steps of `get_manta_sv_by_sample`, based on that `get_manta_sv` now calls `get_combined_sv` instead of `get_merged_results`. I will have to run a few tests tomorrow before it's ready for review. I will call on reviewers when this is done. Thanks!

mattssca:
Testing is done and this PR is ready for review.

R/database.R
```diff
@@ -1215,8 +1176,8 @@ get_combined_sv = function(min_vaf = 0,
 #'
 #' @param min_vaf The minimum tumour VAF for a SV to be returned.
 #' @param min_score The lowest Manta somatic score for a SV to be returned.
-#' @param pass If set to TRUE, include SVs that are annotated with PASS in FILTER column. Default is TRUE.
-#' @param pair_status Use to restrict results (if desired) to matched or unmatched results (default is to return all).
+#' @param pass If set to TRUE, include SVs that are annotated with PASS in FILTER column. Default is FALSE (since sv calls returned with get_combined_sv does not have this column).
```
rdmorin, line 1179: This function always includes these SVs. I think you need to say "exclude SVs that don't contain PASS" to clarify that this leads to dropping of SVs when FALSE. Right now it seems like TRUE brings in things that wouldn't otherwise be retained

R/database.R
```diff
@@ -1215,8 +1176,8 @@ get_combined_sv = function(min_vaf = 0,
...
+#' @param pair_status Use to restrict results (if desired) to matched or unmatched results (default is to return all). Only applies to samples retrieved with get_manta_sv_by_sample, since variant calls from get_combined_sv, does not have this column.
```
rdmorin, line 1180: Why does it not have this column? Doesn't this just come from the metadata and if so, can't you add it?

R/database.R
```diff
@@ -1240,7 +1201,7 @@ get_combined_sv = function(min_vaf = 0,
 #'
 get_manta_sv = function(min_vaf = 0.1,
```
rdmorin, line 1202: Why did you change the default? Do any other functions call this function and, if so, do they assume a certain default?

R/database.R
```diff
@@ -2695,19 +2641,35 @@ get_manta_sv_by_samples = function(these_samples_metadata,
...
-  merged_bedpe = rbind(lifted_calls, no_lift_needed) %>%
-    arrange(CHROM_A, CHROM_B, START_A) %>%
+  merged_bedpe = bind_rows(lifted_calls, no_lift_needed) %>%
     dplyr::select(-need_lift)
-  
+
+  #Deal with chr prefixes based on the selected projection.
+  if(projection %in% c("hg38", "hg19")){
+    merged_bedpe = merged_bedpe %>%
+      dplyr::mutate(CHROM_A = case_when(str_detect(CHROM_A, "chr") ~ CHROM_A, TRUE ~ paste0("chr", CHROM_A))) %>%
```
rdmorin, line 2657: two ifelse instead of two case_when seems a lot simpler for this. I had a hard time following what this does

rdmorin: At the very least say what its doing (i.e. add chr prefix to the chromosome name for builds that expect it, but only add when necessary)

R/database.R
```diff
@@ -2717,15 +2679,16 @@ get_manta_sv_by_samples = function(these_samples_metadata,
...
+#' Please note, if force_lift is set to FALSE, an extra column will be added that states if the returned variant calls need to be lifted. 
+#' The value for this column is returned TRUE (for all rows) if the available genome projection for the selected sample does not match the selected projection (i.e requiring the user to manually lift the calls).
 #'
 #' @param this_sample_id The single sample ID you want to obtain the result from.
-#' @param these_samples_metadata A metadata table containing metadata for this_sample_id, or sample of interest. This parameeter is required.
-#' @param force_lift If TRUE, coordinates will be lifted (if needed) to the selected projection. WARNING! Should always be TRUE if this function is running in stand-alone mode.
+#' @param these_samples_metadata A metadata table containing metadata for this_sample_id, or sample of interest. This parameter is required.
+#' @param force_lift If TRUE, coordinates will be lifted (if needed) to the selected projection. Default is FALSE. WARNING: if running this function in stand-alone mode, set this parameter to TRUE to ensure that the returned calls are in respect to the requested projection.
```
rdmorin, line 2688: What is stand-alone mode? Do you mean "if your code calls this function directly"?

R/database.R
```diff
@@ -2734,16 +2697,19 @@ get_manta_sv_by_samples = function(these_samples_metadata,
 #'
 #' @examples
 #' my_svs = get_manta_sv_by_sample(this_sample_id = "00-14595_tumorD",
-#'                              these_samples_metadata = get_gambl_metadata() %>% dplyr::filter(sample_id == "00-14595_tumorD"))
+#'                                 these_samples_metadata = get_gambl_metadata())
```
rdmorin, line 2700: Do I see an extra bracket?

R/database.R
```diff
@@ -2734,16 +2697,19 @@ get_manta_sv_by_samples = function(these_samples_metadata,
...
 get_manta_sv_by_sample = function(this_sample_id,
                                   these_samples_metadata,
-                                  force_lift = TRUE,
+                                  force_lift = FALSE,
                                   min_vaf = 0.1,
                                   min_score = 40,
                                   pass = TRUE,
-                                  with_chr_prefix = FALSE,
                                   projection = "grch37"){
+
```
rdmorin, line 2709: So this message runs a bunch of times even if this function is called the normal way? Or is the liftover done elsewhere in that case?

R/database.R
mattssca: The reason for this is that `get_combined_sv` does not have a FILTER column. So when this parameter is set to TRUE, we would lose all variants that are returned from `get_combined_sv`. This was a workaround, but something I was hoping to discuss here.

R/database.R
mattssca: Maybe this comment needs to be revisited depending on what we decide to do with the missing FILTER column for `get_combined_sv results` (see my comment below).

R/database.R
mattssca: This has been updated

R/database.R
mattssca: Yes, that is what I meant. I have changed it to be more descriptive.

R/database.R
mattssca: Nope, the example runs just fine:
```
my_svs = get_manta_sv_by_sample(this_sample_id = "00-14595_tumorD",
                                these_samples_metadata = get_gambl_metadata())
```

R/database.R
mattssca: This message only gets printed if the user calls this function directly and with `force_lift = TRUE` (default is FALSE). When this function is internally called by `get_manta_sv_by_samples` this is hard coded to be FALSE, so that all the lifting takes place within that function, on one data frame, instead of running `liftover_bedpe` on all individual flat-files. If that make sense?

rdmorin: Yep. Thanks

R/database.R
rdmorin: changing the default just means you can get the results back but it breaks the ability of the user to filter for only the variants that passed manta's filters. What's not clear to me is whether SVAR has already dropped the non-PASS ones from Manta. If it has, then you can assume that NA is equivalent to PASS. Otherwise, we're a bit stuck. Laura should be able to answer this specific question.

R/database.R
rdmorin: Easiest workaround is to allow PASS or NA to be kept. It would help to know whether we can interpret the combined_SV (with NA) with regard to their PASS status

R/database.R
mattssca: Yes, I was thinking the same thing. Laura confirmed, the variants we're talking about here are all PASS (i.e NA is equivalent to PASS), so it makes this easier. Let me quickly accommodate for this new information in the code, then we should be good to go.

mattssca:
This PR has been updated to solve the reported bug with certain samples not being subject to the liftover step, as intended. This bug was caused by the internal call of `get_manta_sv_by_samples` in `get_manta_sv`. This function was missing the projection parameter, meaning that the selected projection for all samples retrieved with this function defaulted to grch37, resulting in not lifting any samples to any other projection if specified within `get_manta_sv`.

After solving this another issue was discovered. The return for all lifted samples was just an empty data frame. After tweaking the filtering parameters (`pass`, `pairing_status`, `min_vaf`, and `score`) it was discovered that the original lifted data frame actually had samples in it, but that all values for `VAF_tumour` had been swapped to 0, meaning that no samples would show up with default value for `min_vaf` parameter in `get_manta_sv`. This bug was caught in `liftover_bedpe`, which automatically converts all numeric columns to integers, resulting in transforming all `VAF_tumour` values to zero. This code was dropped from this function after confirming with Kostia that this could be done. 

`get_manta_sv` was then tested with a combination of parameters and careful interrogation of samples not returned with `get_combined_sv` (i.e samples returned with `get_manta_v_by_samples`). Coordinates for a set of such samples were compared to make sure that liftover had actually taken place, as well as that chr prefixes had been dealt with in the expected manner. 

In addition, the `with_chr_prefix` parameter for `get_manta_sv` has now been dropped, since all the chr-prefix-handling is now based on the selected projection. If the user still wants to control if their variants are prefixed or not, I think it’s reasonable that they can do so on the return data frame themselves. 

Kostia and I discussed a potential update for how `get_manta_by_sample` and `get_manta_by_samples` work. In short, we reasoned that `get_manta_by_sample` could be updated to not do any lifting, ever. This means this function would instead be interpreted as a helper function, called by `get_manta_by_samples`, where all lifting will occur. This would deprecate the `force_lift` parameter, potentially avoiding any confusion about how this parameter is intended to be used. It also condenses the code for `get_manta_by_sample` to some degree. One downside in doing this would be that a user would no longer be able to call `get_manta_by_sample` on the fly, for specific samples (rare cases?), instead, they would have to depend on `get_manta_by_samples` with the sample ID of interest presented in a metadata data frame, given to the `these_samples_metadata` parameter.

Here’s a reproducible example of how the updates were tested:

```
#run get_manta_sv
all_sv_grch37  = get_manta_sv(projection = "grch37")
all_sv_hg38  = get_manta_sv(projection = "hg38")

> #check if the coordinates are different
> #check a sample with genome_build: hs37d5 (not in return from get_combined_sv)
> dplyr::filter(all_sv_grch37, tumour_sample_id == "816-06-01TD") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A   START_A     END_A CHROM_B   START_B     END_B
1       1 241051430 241051430       1 241132692 241132692
2       1 241070420 241070423       9 104683919 104683922
3      14 106092345 106092346      14 106210781 106210782
4      14 106112860 106112860      14 106211679 106211679

> dplyr::filter(all_sv_hg38, tumour_sample_id == "816-06-01TD") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A   START_A     END_A CHROM_B   START_B     END_B
1    chr1 240888130 240888130    chr1 240969392 240969392
2    chr1 240907120 240907123    chr9 101921637 101921640
3   chr14 105626008 105626009   chr14 105744444 105744445
4   chr14 105646523 105646523   chr14 105745342 105745342

> #check a sample with genome_build: grc37 (not in return from get_combined_sv)
> dplyr::filter(all_sv_grch37, tumour_sample_id == "01-16433_tumorC") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A  START_A    END_A CHROM_B  START_B    END_B
1       2 59584524 59584527       2 63324857 63324860
2       2 63623039 63623042       2 65968099 65968102
3       2 65968014 65968016       2 65968170 65968172
4       2 77795746 77795750       2 77795801 77795801

> dplyr::filter(all_sv_hg38, tumour_sample_id == "01-16433_tumorC") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A  START_A    END_A CHROM_B  START_B    END_B
1    chr2 59357389 59357392    chr2 63097722 63097725
2    chr2 63395904 63395907    chr2 65740965 65740968
3    chr2 65740880 65740882    chr2 65741036 65741038
4    chr2 77568620 77568624    chr2 77568675 77568675

> #check a sample with genome_build: hg38 (not in return from get_combined_sv)
> dplyr::filter(all_sv_grch37, tumour_sample_id == "140127-PL02") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A   START_A     END_A CHROM_B   START_B     END_B
1       1 157369911 157369911       1 157370191 157370191
2       1 242143687 242143687       1 242143756 242143756
3       1 203222147 203222147       8 128779606 128779606
4      11 110090806 110090812      11 111312817 111312823

> dplyr::filter(all_sv_hg38, tumour_sample_id == "140127-PL02") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A   START_A     END_A CHROM_B   START_B     END_B
1    chr1 157400121 157400121    chr1 157400401 157400401
2    chr1 241980385 241980385    chr1 241980454 241980454
3    chr1 203253019 203253019    chr8 127767360 127767360
4   chr11 110220081 110220087   chr11 111442092 111442098

> #check a sample with genome_build: grch37 (is available in get_combined_sv)
> dplyr::filter(all_sv_grch37, tumour_sample_id == "05-11725_tumorA") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A START_A   END_A CHROM_B   START_B     END_B
1       1  948309  948315       6  30661003  30661009
2       1 1308047 1308051       1  50964244  50964248
3       1 1405036 1405051       4  17614779  17614794
4       1 1479720 1479721       6 134845806 134845807

> dplyr::filter(all_sv_hg38, tumour_sample_id == "05-11725_tumorA") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A START_A   END_A CHROM_B   START_B     END_B
1    chr1 1012928 1012935    chr6  30693225  30693232
2    chr1 1372666 1372671    chr1  50498571  50498576
3    chr1 1469655 1469671    chr4  17613155  17613171
4    chr1 1544339 1544341    chr6 134524667 134524669

> #check a sample with genome_build: hg38 (is available in get_combined_sv)
> dplyr::filter(all_sv_grch37, tumour_sample_id == "01-17074T") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A  START_A    END_A CHROM_B  START_B    END_B
1      17 12740288 12740290      17 16931849 16931851
2      19 55960487 55960489       9 98919902 98919904
3      22 29578704 29578707      22 29817673 29817676

> dplyr::filter(all_sv_hg38, tumour_sample_id == "01-17074T") %>% dplyr::select(1:6) %>% head(4)
  CHROM_A  START_A    END_A CHROM_B  START_B    END_B
1   chr17 12836972 12836973   chr17 17028536 17028537
2   chr19 55449121 55449122    chr9 96157621 96157622
3   chr22 29182717 29182719   chr22 29421685 29421687
```

R/database.R
```diff
@@ -1225,14 +1223,24 @@ get_manta_sv = function(min_vaf = 0.1,
   }
 
    if(from_flatfile){
-    all_sv = get_combined_sv(projection = projection)
+    all_sv = get_combined_sv(projection = projection) %>%
+      mutate(FILTER = "PASS") #i.e all variants returned with get_combined_sv() all have PASS in the FILTER column.
```
rdmorin, line 1227: I suggest this mutate step be added into get_combined_sv to make it more self-documenting to users that everything coming from that function already has been selected to meet this criteria. One slightly misleading thing (either way) is that the non-Manta calls coming from get_combined_sv didn't actually "PASS" the same filter, right?

R/database.R
```diff
@@ -2685,10 +2687,10 @@ get_manta_sv_by_samples = function(these_samples_metadata,
 #'
 #' @param this_sample_id The single sample ID you want to obtain the result from.
 #' @param these_samples_metadata A metadata table containing metadata for this_sample_id, or sample of interest. This parameter is required.
-#' @param force_lift If TRUE, coordinates will be lifted (if needed) to the selected projection. Default is FALSE. WARNING: if running this function in stand-alone mode, set this parameter to TRUE to ensure that the returned calls are in respect to the requested projection.
+#' @param force_lift If TRUE, coordinates will be lifted (if needed) to the selected projection. Default is FALSE. WARNING: if your code calls this function directly, set this parameter to TRUE to ensure that the returned calls are in respect to the requested projection.
 #' @param min_vaf The minimum tumour VAF for a SV to be returned. Default value is 0.1.
```
rdmorin, line 2691: What does the function do if you run it without setting force_lift = TRUE and you request a projection that is not available as a file already?

R/database.R
mattssca: Yes, I can move this. Sounds reasonable to me. Without knowing the exact filters for each variant caller, I would not assume so. But I can look into it further if we need to know this.

rdmorin: I don't think that column even exists for non-Manta outputs (i.e. GRIDSS)

R/database.R
mattssca: If this would be done on an individual sample that sample would have an additional column called need_lift and all the rows for that sample set to TRUE, indicating that this sample has not been lifted, and thus the coordinates are not in respect to the selected projection. 

To clarify some potential confusion on how this parameter operates, it is always set to FALSE (default) and this parameter can not be controlled within the by_samples function (hardcodes this parameter to FALSE when internally calling by_sample). This is because anytime you're calling by_samples, liftover will always occur on the subset data frame of sv calls that needs to be lifted (based on the selected projection).

For by_sample, this parameter can be toggled to be TRUE, if so liftover_bedpe will be run on the individual sample. If set to FALSE (default) and the sample in question needs to lifted, an additional column will be added stating this, as detailed above. In addition, if this function is called with this parameter set to TRUE, the user will be suggested to first look for their sample in the desired projection with get_combined_sv.

R/database.R
```diff
@@ -2698,19 +2703,24 @@ get_manta_sv_by_samples = function(these_samples_metadata,
 #' @export
 #'
 #' @examples
-#' my_svs = get_manta_sv_by_sample(this_sample_id = "00-14595_tumorD",
-#'                                 these_samples_metadata = get_gambl_metadata())
-#'
+#' #example 1
+#' #get manta calls for a sample that needs to be lifted to "hg38" and let this function take care of the liftover step for you. 
+#' my_sv = get_manta_sv_by_sample(this_sample_id = "99-27783_tumorA", these_samples_metadata = get_gambl_metadata(), projection = "hg38", force_lift = TRUE)
```
rdmorin, line 2708: Suggestion for Example 2, assuming this works. It shows how the user can either provide just the metadata for their sample of interest or use the combination of parameters.

```
my_metadata = get_gambl_metadata() %>% dplyr::filter(sample_id=="99-27783_tumorA")
my_sv = get_manta_sv_by_sample(these_samples_metadata = my_metadata, projection = "hg38", force_lift = TRUE)
```

mattssca: Added!

rdmorin (approved):

Issue #149, opened 2022-12-29
Allow version-centric use of bundled data


Kdreval:
Current set up of the GAMBLR does not allow for proper tracking and version-centric use of the bundled data.
The user can only see changes through git history, but to actually use one of the earlier versions a separate environment and package installation are required. In addition, with continues development and data addition, some data like aSHM coordinates are bound to change (add new regions, rename existing, curate coordinates etc.). This will not allow reproducible scripts as there is no backward compatibility.

This needs to be resolved by allowing user to use a specific version version of the bundled data on a session-based or project-based basis without need for package installation.

One way this can be addressed is by using a designated and separate data package, which will have incremental version increase as new data is released. At the same time, GAMBLR needs a way to handle this and recognise which version of the bundled data user is requesting at the time of first package load.

Kdreval:
PR with representative set up is at #150

Kdreval:
The bundled data is implemented through the GAMBLR.data package with version control.
Closing this issue as completed.

Issue #147, opened 2022-12-20
Collapse redundant overlapping aSHM regions


rdmorin:
There are redundant aSHM regions in our bundled set of regions. These need to be collapsed to the most representative one for each region to avoid double-counting etc. These should be flagged in the comments of this issue and someone needs to remove the redundant ones. This should really all be done using the "master" list that now lives in [this repo](https://github.com/morinlab/LLMPP/tree/main/resources/curated).

To get things started, here's a few examples:
```
chr_name hg19_start  hg19_end gene    region     regulatory_comment name                  
chr3      186783199 186784291 ST6GAL1 intronic-1 strong_enhancer    ST6GAL1-intronic-1 #fully contained within ST6GAL1-TSS-1    
chr3      186739628 186740875 ST6GAL1 TSS-1      active_promoter    ST6GAL1-TSS-1      
chr3      186781067 186784816 ST6GAL1 intronic-2 strong_enhancer    ST6GAL1-intronic-2 #fully contained within ST6GAL1-TSS-2 
chr3      186737482 186741455 ST6GAL1 TSS-2      strong_enhancer    ST6GAL1-TSS-2     
```

rdmorin:
There is one larger region for PAX5 that is redundant with two smaller ones and should be removed (indicated below).
```
chr_name hg19_start hg19_end gene  region            regulatory_comment              name          size
chr9       37023396 37027663 PAX5  intron-1          intronic                        PAX5-intron…  4267
chr9       37369209 37372160 PAX5  distal-enhancer-1 enhancer                        PAX5-distal…  2951
chr9       37382267 37385854 PAX5  distal-enhancer-2 enhancer                        PAX5-distal…  3587
chr9       37395932 37409239 PAX5  distal-enhancer-3 enhancer                        PAX5-distal… 13307 
chr9       37032576 37037704 PAX5  TSS-1             active_promoter                 PAX5-TSS-1    5128
chr9       37021323 37043181 PAX5  TSS-2             active_promoter-strong_enhancer PAX5-TSS-2   21858 #remove this
```

rdmorin:
This is a trickier one to consolidate. Removing the larger region that encompasses the two others would mean no coverage for one place that is commonly mutated. Perhaps this large region should be split into three or four consecutive regions?
```
  chr_name hg19_start hg19_end gene   region   regulatory_comment name             size
chr9       37192080 37207549 ZCCHC7 intron-3 intronic           ZCCHC7-intron-3 15469
[...]

PR #150, opened 2022-12-29
Handling of bundled data through designated separate repo

Kdreval:
This is related to issues #147 and #149 
Currently this set up supports use of ashm coordinates of specific version as GAMBLR is first loaded. The version is stored in config and can be modified on session-by-session, project-by-project basis.
If this set up proves to be efficient, it will be extended to handle also lymphoma genes and other bundled GAMBLR data.


R/load_data.R
```diff
@@ -0,0 +1,96 @@
...
+    versions <- sort(
+        numeric_version(
+            versions
+            )
+    )
+
+    latest_version <- max(versions)
+
+    # Which version did the user requested in config?
+    requested_version <- config::get("bundled_data_versions")[[mode]]
```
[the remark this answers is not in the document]
Kdreval: Right! I'll fix this!

config.yml
```diff
@@ -160,3 +160,7 @@ default:
                 hg38-nci: "04-24937N-Schmitz"
                 hg38: "BLGSP-71-06-00286-99A-01D"
                 hg19-clc: "PA011-G"
+
```
rdmorin, line 163: Can we also use this to specify a default metadata for example data sets bundled with the package?

Kdreval: Yes, I was thinking of it the same way as well - and if the concept in this draft is something we will move forward with I will switch the metadata (+ ssm, SV, CNV) to the helper package as well

R/load_data.R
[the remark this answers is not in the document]
Kdreval: When I started adding other objects and looking at this, I realized here that we would have this in plural - because different data types develop more than the others and iterate through versions differently. For example, for aSHM sites we already have versions 0, 1, 2, and 3 - but for lymphoma genes only 0 and 1. So config here will have different keys depending on the data type.

config.yml
Kdreval: I have also omitted from this PR bundling the example metadata and datasets - I think this can be it;s own dedicated PR since we would also want to generate examples for each function using that example data and for metadata we also need to work out minimal required columns

R/load_data.R
```diff
@@ -0,0 +1,106 @@
+# Helper functions not for export
+
+#' Check for a version of data to load
```
mattssca, line 3: Consider adding a descriptive title for this function. Currently, this function has the title "Check for a version of data to load". The first line of the function doc, is where the title is extracted from.

R/load_data.R
```diff
@@ -0,0 +1,106 @@
...
+#'
+#' This function determines if a user is requesting the latest version
```
mattssca, line 5: This is where I would specify that this is a helper function (to have it consistent with other helper functions).

R/load_data.R
```diff
@@ -0,0 +1,106 @@
...
+#' of the bundled data or wants to access one of the earlier veresions.
+#'
+#' @param mode Determines which data to handle. Defaults to somatic_hypermutation_locations. Will grow with more options as more data is version tracked.
+#' @param this_genome_build The genome build of the data if coordinate-based. Accepts grch37 (default) or hg38.
+#'
+#' @return data frame
+#' @import config dplyr readr GAMBLR.data
+#'
```
mattssca, line 13: I have gone over all the GAMBLR helper functions on my branch. This was done as a step in preparing the documentation for building a site from the source code with `build_site` from `pkgdown`. This function takes all `.Rd` files that live in the man folder and generates function-specific HTML files. This is not something we want for helper functions (since they are not exported in NAMESPACE). 

The fix for this is to add `@noRd` in the documentation for such functions. This prevents the `.Rd` file from being generated. 

I think you should also specify this for this function to keep things consistent. Let me know if this doesn't make sense, or if you need me to further clarify. 

Thanks,

mattssca (changes requested):
Thanks for a great update Kosta. I have a few comments related to how you handle the function documentation. Let me know if these make sense to you. Thanks!

mattssca:
Thanks for the updates Kostia, for some reason Git didn't notify me on email that you've updated this branch. 

However, I cloned this branch to test the newly added `collate_pga` and was given an error in return when running the example in the docs. This is the error message that was returned:

```
> pga_metrics <- collate_pga(these_samples_metadata = meta)
Collating the PGA results ...

Error in UseMethod("filter") : 
no applicable method for 'filter' applied to an object of class "function" 
  
5. filter(., start <= arm_end & arm_start <= end)

4. arrange(., sample, chrom, start)

3. df %>% filter(start <= arm_end & arm_start <= end) %>% arrange(sample, 
    chrom, start) at utilities.R#3886
    
2. calculate_pga(this_seg = multi_sample_seg) at utilities.R#3726

1. collate_pga(these_samples_metadata = meta) 
```
I am not sure if this error is related to `dplyr` not being specified at the call of `filter` (line [3886](https://github.com/morinlab/GAMBLR/pull/150/files#diff-de996a5b176cafa3ba2b50ececb35f5c17d12b34ae07a5f2eaeb6fadff2301c8R3887)) or if the reason is that the `df`(same line) is previously undefined, or if it's something completely different. But you might want to look into this. Let me know if you need any help testing.

Thanks,

mattssca:
Thanks Kostia! I will clone this branch and try it out!

mattssca (changes requested):
Thanks for the update Kostia, it sure feels nice to finally have capture support for the CN functions. I have some comments before I approve this PR. Let me know if there is anything you want me to clarify. Thanks again, 

1. There are also a few `fancy_x_plot` functions (`fancy_v_chrcount`, `fancy_snv_chrdistplot`, `fancy_v_count`, `fancy_v_sizedis`, `fancy_ideogram`, `fancy_circos_plot`) that are internally calling `assign_cn_to_ssm`. These functions should also have the recently added `this_seq_type` parameter added.

2. Other functions that also internally call `assign_cn_to_ssm` are; `estimate_purity`, `copy_number_vaf_plot`, (these functions should also have the new parameter added).

3. The snakefile used for retrieving data (to run GAMBLR remote) should probably also be updated to fetch the capture version of the merged CNV data file. Otherwise, the remote functionality will be broken whenever seq type capture is requested.

4. As a side note, I ran my recently developed `get_missing_from_merge` with `merge = "cnv"` and `this_seq_type = "capture"` and 1 sample appears to be missing (1669-06-05PD). Is this expected?

5. The first example in `get_cn_states` does not seem to be working for me. I am interested if this works for you. I've tried this example on my own branch and on the recently cloned Kostia branch, but errors out in both cases. This is the example I am referring to `cn_matrix = get_cn_states(regions_bed = grch37_lymphoma_genes_bed)`

6. I tested out all the affected functions with different parameters to get a good sense of how they can handle this update. The results look great, with one exception. `assign_cn_to_ssm` errors out when capture is specified as the seq type. What about adding a working example for this function with the seq type set to capture? Or detail the limitations of the function in the docs. What do you think?

7. `collate_pga` now works as intended, nice! I am wondering, do you want to update the cached files (for `collate_results`) as well? I checked, and PGA seems to missing from these results. If not, this can probably be done in the near future.

Thanks!

Kdreval:
Thanks Adam! Here is what I think:

1. These have been added and documentation updated/regenerated :white_check_mark: 

2. Added the new parameter to `estimate_purity`, `copy_number_vaf_plot`, and updated the documentation :white_check_mark: 

3. I don't have local R/Rstudio and could not test the remote functionality. Could you help with this if you have the remote functionality setup and have testing case? :white_flag: 

4. Yes, this is expected. That sample is problematic and will be eventually kicked out of GAMBL. [Here](https://morinlabsfu.slack.com/archives/CUCMSL2N8/p1673483354753709) are more details.  :white_check_mark: 

5. I noticed this example was removed too - yes. It does generates an error for me too when I test it as is - but it is not capture related or cnv related. The reason for this is that the region_bed argument supplied in this example contains a corrupted entry:
```
171 7                               151832010    152133090 KMT2C      
172 8                                 1993155      2113475 MYOM2      
173 8   8640864 8751155 MFHAS1             NA           NA NA         
174 8                                20054878     20084330 ATP6V1B2   
175 8                                35092975     35654068 UNC5D   
```
It of course is not producing a desired output (column in the matrix) and therefore there is mismatch in the length of names and columns - so the function errors. I created an issue related to this #193 and this can be addressed in a separate PR. :white_flag: 

6. That function is broken in a sense that it only works with battenberg outputs and needs a revamp on how it works. I have also created issue #194 and it can be addressed separately. :white_flag: 

7. Sure, I can update the flatfiles as well - only thinking it should be done after PR is merged so the code how those files are generated is on master and accessible to everyone/reproducible. :white_flag:

mattssca:
Thanks for answering my questions Kostia. I can indeed test and address the remote functionality related to capture CNVs. I'll add support for this in my next PR. I can also look into the other outstanding items that you created issues for. I think this PR looks good, I'll approve the changes.

mattssca (approved):

Kdreval:
Regarding the Question 5 above, I have resolved it and closed the issue #193 . I will wait till the GitHub Actions pass on the latest commit for this branch and will then merge it to master. Thanks Adam!

Issue #161, opened 2023-01-25
sample_id parameter should be renamed to this_sample_id or these_sample_ids


mattssca:
Functions that are still calling `sample_id` parameter, should receive an update to rename this parameter to either `this_sample_id` or `these_sample_ids` so that this parameter name is consistent throughout all functions in GAMBLR. 

These are the functions that are currently using `sample_id`:

```
get_combined_sv
plot_multi_timepoint
finalize_stud (in portal.R)
assemble_file_details
estimate_purity
make_igv_snapshot
cnvKompare
```

Issue #160, opened 2023-01-20
Conflicting function (margin) from randomForest and ggplot2.


mattssca:
This conflict was found when regenerating vignettes. Here's an example of an instance where a plotting function fails, due to this conflict.

```
> #set region for the plot to highlight.
> mybed = data.frame(start = 128747680,
+                    end = 128753674,
+                    name = "MYC")
> #full region to be plotted.
> region = "chr8:128737680-128763674"
> #get gambl metadata for all FL and DLBCL samples.
> fl_dlbcl_metadata = get_gambl_metadata() %>%
+   dplyr::filter(pathology %in% c("FL", "DLBCL"))
                                                                                                                                                                                                                                                                
> #get SSm data for the specified region.
> my_mutations = get_ssm_by_region(region = region)
                                                                                                                                                                                                                                                                
> #construct plot
> ashm_rainbow_plot(mutations_maf = my_mutations,
+                   drop_unmutated = TRUE,
+                   metadata = fl_dlbcl_metadata,
+                   hide_ids = TRUE,
+                   bed = mybed,
+                   region = region,
+                   classification_column = "pathology",
+                   custom_colours = c("FL" = "#EA8368", "DLBCL" = "#479450"))
not using the database
 Error in UseMethod("margin") : 
  no applicable method for 'margin' applied to an object of class "c('double', 'numeric')" 
```

The fix is easy, for ggplots calling the `margin` function, specify that this is from ggplot as such `ggplot2::margin()`.

I will self-assign this issue. I am posting this here for tracking purposes.

rdmorin:
I can't see from what you posted where the issue arose. Is it an automatic merge somewhere in the function that's being called?

mattssca:
Not sure when this issue arose. I think `randomForest` was added sometime in early December (imported by `classify_fl`) and my guess is that the two plotting functions that are using the margin function from ggplot2 (`ashm_multi_rainbow_plot` and `ashm_rainbow_plot`) has not been used since this package was added.

rdmorin:
I follow you now. I thought you meant this had to do with a merge conflict. It's just a namespace conflict. Makes sense that it came along with the addition of that package.

Kdreval:
The randomForest dependency was removed by moving out the dependent functionality to a separate repo. There will be no more conflicts of randomForest functions 😃

Issue #159, opened 2023-01-20
Remote support for get_sample_cn_segments is broken


mattssca:
There's a bug in this function that forces the function to use the path to the remote file instead of the local one.

`full_cnv_path =  paste0(config::get("project_base", config = "default"), cnv_path)`

Getting rid of `config=default` seems to work, but testing has to be made to ensure it does not break anything else.

Issue #157, opened 2023-01-19
Update Readme


rdmorin:
The documentation in the Readme (beginning with the first section in getting started ) is out of date. New users are asked to set up the database but we've moved away from that. Can we please pare this down and ensure only the current/relevant/necessary information is in the setup guide?

PR #164, opened 2023-01-27
GAMBLR Improvements and Bugfixes

mattssca:
This PR is providing a fix for many of the active issues on GAMBLR, as well as other quality-of-life improvements. The following issues have been addressed in this PR:

- https://github.com/morinlab/GAMBLR/issues/161
- https://github.com/morinlab/GAMBLR/issues/160
- https://github.com/morinlab/GAMBLR/issues/159
- https://github.com/morinlab/GAMBLR/issues/157
- https://github.com/morinlab/GAMBLR/issues/156
- https://github.com/morinlab/GAMBLR/issues/153
- https://github.com/morinlab/GAMBLR/issues/128
- https://github.com/morinlab/GAMBLR/issues/102

- This PR also includes updates that are ensuring that the thematic vignettes can be knitted in remote sessions as well.
- Many updates in this PR are related to the function documentation overhaul.
- All functions that are using the config in any way to get file paths have been updated to encapsulate these statements in the newly added helper function that checks the config keys.
- README.md has been updated with a new section called "For Developers". This section details what the best practices are for designing and adding a new function to GAMBLR, complete with examples and a function template that can be used for designing new functions as well as an example of an adequately documented GAMBLR function.
- Minor hot fixes prevented the vignettes from being knitted. For some reason, the vignettes errors out when a warning gets thrown. Even if the `warnings` in the chunk option is set to FALSE.
- This branch was tested by running the vignettes as well as the [test_funcitons.R](https://github.com/morinlab/GAMBLR/blob/master/test_functions.R) script.
- the vignettes have also been updated to rely on modified code.
- The package documentation has also been re-generated to match all the updates included in this PR.

# Pull Request Checklists

### Required

- [x] I updated and/or successfully knitted a vignette that relies on the modified code (which ones?)

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


R/annotation.R
```diff
@@ -213,14 +260,21 @@ annotate_igh_breakpoints = function(annotated_df,
...
+#' For information on how to use the additional parameters, refer to the parameter descriptions and function examples. 
 #'
-#' @param maf_df Data frame of MAF-format mutations (default is to retrieve automatically).
+#' @param maf_df Data frame of MAF-format mutations.
 #' @param lymphoma_type Optional keyword to find genes for annotation (e.g. "BL","DLBCL","FL") is not specifying driver genes.
 #' @param driver_genes Optional vector of Hugo_Symbol of genes for coding mutations.
 #' @param noncoding_regions Optional named vector of regions to use to further restrict noncoding mutations per gene.
 #'
-#' @return Driver ssm kepts from maf.
+#' @return Driver ssm kept from maf.
```
rdmorin, line 275: Is the maf altered in any way? Are there any new columns?

R/checks.R
```diff
@@ -17,26 +21,38 @@ get_template_wildcards = function(parent_key,template_key){
...
+#' 
+#' @description Check for a remote session and automagically confirm setup will work properly.
 #' 
-#' This function determines if a user is working in GAMBLR remotely and, if so, will
+#' @details This function determines if a user is working in GAMBLR remotely and, if so, will
 #' check if their config is loaded properly and ssh_session is available. 
 #'
 #' @param auto_connect Set to TRUE to ensure an ssh_session is created if absent
 #'
 #' @return
```
rdmorin, line 47: Does it return anything? I think maybe a boolean TRUE/FALSE depending on whether there is a remote connection?

R/checks.R
```diff
@@ -161,19 +187,26 @@ check_file_details = function(relative_paths){
...
 #'
 #' @param compare_timestamps Whether the function will compare timestamps on your local files to the remote copy. Only relevant if you are working remotely.
 #' @param ssh_session The ssh_session object see get_ssh_session() for more information. Only relevant if you are working remotely.
 #' @param archive_mode This is not currently working but the idea here is to keep a GSC archive of GAMBL in sync with the actively updated outputs
-#' @param force_backup 
+#' @param force_backup Boolean parameter set to FALSE per default.
 #'
 #' @return
+#' 
+#' @import dplyr tidyr
```
rdmorin, line 204: would adding glue to this mean we don't have to put glue::glue everwhere?

R/checks.R
```diff
@@ -383,3 +426,24 @@ check_expected_outputs = function(tool_name="battenberg",seq_type_filter="genome
 }
 
 
+#' @title Check Config Value.
+#' 
+#' @description Check the existence of a specific config key. 
+#' The function will notify the user if no such key exists.
```
rdmorin, line 432: "will notify the user and end the program if..."

R/database.R
```diff
@@ -4,18 +4,24 @@ cnames = c("CHROM_A", "START_A", "END_A", "CHROM_B", "START_B", "END_B", "NAME",
...
 #'
-#' @param tool_name The tool or pipeline that generated the files (should be the same for all).
+#' @description Exclude samples that have been excluded from certain analyses and drop from merges.
 #'
-#' @return
+#' @details Specify the tool or pipeline responsible for generating the files with `tool_name` and this function will return a list of excluded sample IDs.
+#'
+#' @param tool_name The tool or pipeline that generated the files (should be the same for all). Default is "slms-3".
+#'
+#' @return A list of sample IDs.
```
rdmorin, line 15: I've seen you refer to a vector as a "list" many times. There is an important distinction between the two (at least in R) so please be sure you describe variables using the right class name. It should be "a vector of sample IDs" and that is probably almost always what you mean when the documentation says "list"

R/database.R
```diff
@@ -32,30 +38,38 @@ get_excluded_samples = function(tool_name = "slms-3"){
...
+#' This is done either by combining the contents of individual MAF files or subset from a merged MAF (wraps get_ssm_by_samples).
+#' In most situations, this should never need to be run with `subset_from_merge = TRUE`. Instead use one of `get_coding_ssm` or `get_ssm_by_region`
 #'
 #' @param these_patient_ids A vector of sample_id that you want results for. This is the only required argument.
-#' @param these_samples_metadata Optional metadata table
-#' @param tool_name Only supports slms-3 currently
-#' @param projection Obtain variants projected to this reference (one of grch37 or hg38)
-#' @param flavour Currently this function only supports one flavour option but this feature is meant for eventual compatibility with additional variant calling parameters/versions
-#' @param min_read_support Only returns variants with at least this many reads in t_alt_count (for cleaning up augmented MAFs)
+#' @param these_samples_metadata Optional metadata table.
```
rdmorin, line 50: these_patient_ids A vector of sample_id that you want results for. 
The above argument should not be required if these_samples_metadata is provided. This is how every GAMBLR function should work. If the ids vector is missing, the function should get it from the metadata as long as the user provided it. 

these_samples_metadata Metadata table. It is expected the user has subset this to contain the rows corresponding to the patients of interest.

R/database.R
```diff
@@ -112,36 +126,40 @@ get_ssm_by_patients = function(these_patient_ids,
...
+#' Instead use one of `get_coding_ssm` or `get_ssm_by_region`.
+#' See `get_ssm_by_sample` for more information.
 #'
 #' @param these_sample_ids A vector of sample_id that you want results for.
-#' @param these_samples_metadata Optional metadata table
-#' @param tool_name Only supports slms-3 currently
-#' @param augmented default: TRUE. Set to FALSE if you instead want the original MAF from each sample for multi-sample patients instead
-#' @param projection Obtain variants projected to this reference (one of grch37 or hg38)
-#' @param flavour Currently this function only supports one flavour option but this feature is meant for eventual compatability with additional variant calling parameters/versions
+#' @param these_samples_metadata Optional metadata table.
```
rdmorin, line 139: As in my last comment, please confirm the function works with either these_sample_ids or these_samples_metadata such that either of them can be provided to ensure the function runs only on the IDs the user wants. As long as this is how it works, please update the documentation so this is clear.

rdmorin (changes requested):

NAMESPACE
```diff
@@ -116,27 +113,39 @@ import(RColorBrewer)
 import(RMariaDB)
 import(S4Vectors)
 import(SRAdb)
-import(broom)
```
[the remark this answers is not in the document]
mattssca: This mistake was most likely due to an auto-merge, where `prettyForestPlot` had two import fields. I have removed the redundant @import section from this function (the one without broom specified) and re-generated the package docs. GAMBLR now imports broom, as it should. Good catch!

R/annotation.R
```diff
@@ -2,24 +2,45 @@
 coding_vc = c("Frame_Shift_Del", "Frame_Shift_Ins", "In_Frame_Del", "In_Frame_Ins", "Missense_Mutation", "Nonsense_Mutation", "Nonstop_Mutation", "Splice_Region", "Splice_Site", "Targeted_Region", "Translation_Start_Site")
 
 
-#' Annotate and auto-drop a MAF data frame with existing blacklists to remove variants that would be dropped during the merge process.
+#' @title Annotate SSM with Blacklists
 #'
-#' @param mutations_df df with mutation data.
```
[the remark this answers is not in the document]
mattssca: Got it! This has ben updated.

R/annotation.R
```diff
@@ -2,24 +2,45 @@
...
-#' @param unix_group The unix group (should be the same for all)
-#' @param tool_name The tool or pipeline that generated the files (should be the same for all)
-#' @param flavour Set to "clustered" if you want to use the blacklist for the new and improved SLMS-3 outputs (otherwise leave empty)
-#' @param genome_build The genome build projection for the variants you are working with (default is grch37)
-#' @param project_base Optional: A full path to the directory that your blacklist_file_pattern is relative to
-#' @param blacklist_file_pattern Optional: A string that contains the relative path to your blacklist file from after the project_base (i.e. results) with any wildcards surrounded with curly braces
-#' @param drop_threshold The minimum count from one of the blacklists to drop a variant
-#' @param verbose For debugging, print out a bunch of possibly useful information
-#' @param invert USE WITH CAUTION! This returns only the variants that would be dropped in the process (opposite of what you want, probably)
+#' @description Annotate and auto-drop a MAF data frame with existing blacklists.
```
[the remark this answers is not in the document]
mattssca: It's because I added punctuation at the end of every parameter description.

R/annotation.R
```diff
@@ -2,24 +2,45 @@
...
 #' @return A MAF format data frame with two new columns indicating the number of occurrences of each variant in the two blacklists.
+#' 
+#' @import dplyr glue readr tidyr
+#' 
 #' @export
 #'
-#' @examples deblacklisted_maf_df = annotate_ssm_blacklist(original_maf_df)
+#' @examples 
+#' #get SSMs
+#' original_maf_df = get_coding_ssm()
```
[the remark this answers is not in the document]
mattssca: Yes, that makes sense. The example has been updated in this regard.

R/annotation.R
```diff
@@ -95,16 +124,23 @@ annotate_ssm_blacklist = function(mutations_df,
...
-    additional_blacklist = glue(full_path) %>% read_tsv()
+    repo_base = check_config_value(config::get("repo_base"))
+    full_path = paste0(repo_base, check_config_value(config::get("resources")$curated_blacklist))
+    
+    additional_blacklist = glue::glue(full_path) %>% 
+    readr::read_tsv()
+
     additional_blacklist = additional_blacklist %>%
-      separate(chrpos, into = c("Chromosome", "Start_Position"), sep = ":") %>% mutate(Start_Position = as.numeric(Start_Position))
+      separate(chrpos, into = c("Chromosome", "Start_Position"), sep = ":") %>%
```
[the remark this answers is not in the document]
mattssca: `separate` is from `tidyr`.

R/annotation.R
```diff
@@ -124,13 +160,17 @@ annotate_ssm_blacklist = function(mutations_df,
 }
 
 
-
```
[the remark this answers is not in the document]
mattssca: I would like you (or anyone else) to explain this function. Currently, it does not do anything, thus the return has been declared as "Nothing". Alternatively, I could remove the export for this function so that it doesn't populate the NAMESPACE.

R/annotation.R
mattssca: There are no new columns, only a subset of the rows (mutations) identified as putative driver mutations. I have updated the return to state this clearly and expanded the example for the same function.

R/checks.R
mattssca: Yes, I have updated the return description for this function (it actually returns the Boolean for `check_host`).

R/checks.R
mattssca: Yes, based on my testing. But I have not applied these changes globally. 

There are basically two ways to tell a function what packages are needed. First, tell the function under the import section (in the documentation). Or, directly specify the package when a function from an external package is being called, followed by double colons (i.e `package::function`). Both options are valid and should achieve the same thing, at least from the perspective of the main function. Not sure how it would affect the overall documentation. I am thinking that the second option would mean that the external package would not show up under imports in the NAMESPACE.

R/checks.R
mattssca: Nice, I have updated the docs for this function.

R/database.R
mattssca: Right, good point. I have gone through the complete package and updated this where needed.

R/database.R
mattssca: Yes, 100%. I have updated the parameter descriptions, function details and examples to make this clear. The function works as advertised with the two parameters discussed.

R/database.R
mattssca: I can confirm it works as advertised. I have also updated the documentation to make this clear as well as added an additional example demonstrating how to use this function with an already subset metadata table.

NAMESPACE
```diff
@@ -121,22 +117,33 @@ import(circlize)
...
 import(metaviz)
 import(parallel)
 import(plotly)
+import(purrr)
+import(readr)
+import(reshape2)
 import(rtracklayer)
-import(tidyverse)
+import(ssh)
+import(stats)
```
Kdreval, line 142: which package imports stats? We should drop this because it introduces conflicts with function names and gives lots of warnings on package installation/importing

NAMESPACE
```diff
@@ -121,22 +117,33 @@ import(circlize)
...
 import(maftools)
 import(metaviz)
 import(parallel)
 import(plotly)
+import(purrr)
+import(readr)
+import(reshape2)
 import(rtracklayer)
-import(tidyverse)
+import(ssh)
```
Kdreval, line 141: I might misremember this - but I think there was some issue with installing the `ssh` package on GSC. So we opted to not import it and leave it to remote users to take care of. Maybe I am not correct or this has been resolved - but I would keep this out of the dependency list

NAMESPACE
```diff
@@ -121,22 +117,33 @@ import(circlize)
...
+import(glue)
 import(grid)
+import(gridExtra)
 import(maftools)
 import(metaviz)
 import(parallel)
 import(plotly)
+import(purrr)
+import(readr)
+import(reshape2)
```
Kdreval, line 139: Which function imports reshape2? I just dropped it in the latest mster update but seems like it is used somewhere else, too?

NAMESPACE
```diff
@@ -79,19 +79,15 @@ export(intersect_maf)
 export(liftover_bedpe)
 export(maf_to_custom_track)
 export(make_igv_snapshot)
-export(plot_multi_timepoint)
```
Kdreval, line 82: Not sure what happened with these 4 functions that were discontinued - but trust it was discussed somewhere. Just want to double-check with you it is intentional

NAMESPACE
```diff
@@ -121,22 +117,33 @@ import(circlize)
 import(config)
 import(cowplot)
 import(data.table)
-import(dbplyr)
 import(digest)
 import(dplyr)
+import(forcats)
```
Kdreval, line 122: which function was using forcats? Maybe we can rework it and avoid introducing this as dependency?

NAMESPACE
```diff
@@ -121,22 +117,33 @@ import(circlize)
...
 import(g3viz)
 import(ggbeeswarm)
 import(ggplot2)
 import(ggpubr)
 import(ggrepel)
 import(ggsci)
 import(ggthemes)
+import(glue)
 import(grid)
+import(gridExtra)
```
Kdreval, line 132: What is using gridExtra? Maybe we should rework that function to avoid another dependency? It is also probably already a sub-dependency of one of the packages like Complexeatmap or circlize

R/annotation.R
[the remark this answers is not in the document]
Kdreval: Did you do it by hand or there are some scripts/functions? I think it might not be sustainable in a long-term future unless we find a way to consistently follow this. It might be that it is now consistently introduced everywhere but the new functions will still be missing these so will then end up with a need another review/PR and it will keep going on like this. Just thinking if we can automate the things like punctuation.

R/annotation.R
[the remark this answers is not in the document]
Kdreval: You can compare the dimensions of the data before and after this function is run. Note: to see what it is doing you need to get the maf that has not been deblacklisted before, otherwise your input will be identical to the output since this functionality is used in [lcr-modules](https://github.com/LCR-BCCRC/lcr-modules/blob/78c3ea60e0b8cc2c27595b6dc7041aff2b71d9f9/modules/vcf2maf/1.3/vcf2maf.smk#L221-L244) to deblacklist mafs at the first place.

NAMESPACE
mattssca: Agreed. Assuming you're meaning "what function* imports stats", these are the functions that are currently importing stats;

```
FtestCNV
prettyForestPlot
prettyOncoplot
```

NAMESPACE
mattssca: Right, thanks for pointing this out. I will make the changes necessary

NAMESPACE
mattssca: Yes, it's imported by `prettyForestPlot`.

NAMESPACE
mattssca: Yes, these functions are still in GAMBLR, I just removed the @export for them, since they are all either in draft mode or/and are not yet doing anything. I thought this was the best approach since it does not make sense for a user to have access to functions that are not doing anything. Especially since this package is now open to the public. I also talked this over with Ryan when he was at BC Cancer a week ago, and he agreed. Thanks for checking!

NAMESPACE
mattssca: Currently, the only package that is importing `forcats` is `prettyForestPlot`.

NAMESPACE
mattssca: Currently, there's only one package that imports gridExtra and it is `comp_report`. It's the wrapper function for all the sample-level fancy plots. It uses `gridExtra` to arrange all plots and output a pdf. I can look into if the same output can be achieved without the dependency on this package.

NAMESPACE
Kdreval: I think your branch is not up to date with master. This was re-worked to remove the reshape2 dependency. The same also applies to stats package - in the recent master it is no longer imported

NAMESPACE
Kdreval: Thanks for confirming!

R/utilities.R
```diff
@@ -2553,26 +2885,30 @@ get_gambl_colours = function(classification = "all",
...
+#' 
+#' @details Returns a list with BAM paths for tumour, normal and mrna data.
+#' This function expects a sample ID (`sample`) or a patient ID (`patient`).
 #'
 #' @param sample Either provide sample_id or patient_id.
 #' @param patient Either provide sample_id or patient_id.
 #'
 #' @return A list that contains the genome_build and an igv-friendly build (igv_build), a list of bam file paths for tumour, normal and mrna data.
+#' 
+#' @import dplyr utils
```
Kdreval, line 2900: The utils is a default R package always available when R starts. Do we need to explicitly import it here? Also, couldn't see where it is used below

R/viz.R
```diff
@@ -359,10 +364,15 @@ prettyRainfallPlot = function(this_sample_id,
   return(p)
 }
 
+
 gene_mutation_tally = function(maf_df,these_samples_metadata,these_genes,grouping_variable="cohort"){
   meta = dplyr::select(these_samples_metadata,sample_id,{{grouping_variable}})
+  maf_filt = dplyr::filter(maf_df,Hugo_Symbol %in% these_genes, Variant_Classification %in% coding_class) %>%
   maf_filt = dplyr::filter(maf_df,Hugo_Symbol %in% these_genes, Variant_Classification %in% coding_class) %>%
     dplyr::filter(Variant_Classification !="Silent")
+  meta_anno = left_join(maf_filt,meta,by=c("Tumor_Sample_Barcode"="sample_id")) %>%
```
Kdreval, line 373: This seems to be a duplicate of what is below. Wonder if this is intentional?

R/viz.R
```diff
@@ -383,17 +399,30 @@ gene_mutation_tally = function(maf_df,these_samples_metadata,these_genes,groupin
 #' @param colour_index Optional named character vector with a name for each gene in these_genes and a colour as the value
 #'
 #' @return data frame with counts for each gene
+#' 
+#' @import wordcloud RColorBrewer dplyr utils
```
Kdreval, line 403: Same here for utils, wonder if it should be explicitly mentioned or  we can avoid it?

R/viz.R
```diff
@@ -4823,66 +5106,83 @@ fancy_circos_plot = function(this_sample,
...
-                            hide_legend = FALSE,
-                            chr_select = paste0("chr", c(1:22)),
-                            plot_title = paste0(this_sample),
-                            plot_subtitle = paste0("SV sizes for Manta calls. Dashed line annotates mean variant size.\nVAF cut off: ", vaf_cutoff,", SV size cut off: ", size_cutoff),
-                            projection = "grch37"){
+#' myplot = fancy_sv_sizedens(this_sample_id = "HTMCP-01-06-00422-01A-01D")
+#' myplot2 = fancy_sv_sizedens(this_sample_id = "HTMCP-01-06-00422-01A-01D", size_cutoff = 0, chr_select = c("chr1", "chr2"))
+#'
+fancy_sv_sizedens = function(this_sample_id,
+                             maf_data,
```
Kdreval, line 5171: Seems like this is just space additions?

Kdreval (commented):
Thanks, Adam, so much work and important updates! I had some other questions/suggestions

NAMESPACE
Kdreval: The `stats` package is a default R package and is available when R starts. Therefore, to use `p.adjust`, we do not need to refer to the package specifically. Including the `stats`  in NAMESPACE explicitly only created name conflicts and many warnings generated once GAMBLR is loaded.

NAMESPACE
mattssca: For this, the code is still using functions from the `stats` package (from my testing).

- FtestCNV calls `fisher.test` and `adjust.p`
- prettyForestPlot calls `adjust.p`
- prettyOncoplot calls `setNames`

NAMESPACE
mattssca: Ok nice! Didn't know that `ggpubr` could achieve the same thing. I'll leave it as is for now.

R/checks.R
```diff
@@ -90,18 +112,21 @@ check_host = function(auto_connect=FALSE,verbose=FALSE){
...
 
-check_times = function(relative_paths,archive_mode=FALSE,force_backup=FALSE){
+
+check_times = function(relative_paths,
+                       archive_mode = FALSE,
+                       force_backup = FALSE){
   
-  local_base = base_path=config::get("project_base")
-  remote_base = base_path=config::get("project_base",config="default")
+  local_base = base_path = check_config_value(config::get("project_base"))
```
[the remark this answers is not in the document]
mattssca: For this, I just added the newly added `check_config_value` to any **existing** line that gets paths from the config. So, if this indeed is something funny going on here, it would be outside the scope of what's introduced in this PR. But should be addressed nonetheless.

R/checks.R
```diff
@@ -90,18 +112,21 @@ check_host = function(auto_connect=FALSE,verbose=FALSE){
...
+  remote_base = base_path = check_config_value(config::get("project_base",config="default"))
   if(archive_mode){
     if(local_base == remote_base){
       message("checking against local archive")
     }else{
       message("Currently, this mode must be run on the GSC (not remotely)")
       return(NULL)
     }
-    local_base = base_path=config::get("archive")
+    local_base = base_path = check_config_value(config::get("archive"))
```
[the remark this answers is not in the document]
mattssca: Same here :)

R/database.R
```diff
@@ -596,10 +637,11 @@ get_gambl_metadata = function(seq_type_filter = "genome",
   sample_meta = sample_meta %>%
     dplyr::filter(seq_type %in% seq_type_filter & tissue_status %in% tissue_status_filter) %>%
     dplyr::select(-sex)
+  
   #if only normals were requested, just return what we have because there is nothing else to join
-  if(tissue_status_filter == "normal"){
-    return(sample_meta)
-  }
+  #if(tissue_status_filter == "normal"){
```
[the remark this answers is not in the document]
mattssca: Yes, this solved a previous warning message that prevented the vignettes to knitted.

R/database.R
```diff
@@ -1201,57 +1253,83 @@ get_combined_sv = function(min_vaf = 0,
...
+                        these_samples_metadata,
+                        projection = "grch37",
                         chromosome,
                         qstart,
                         qend,
                         region,
-                        from_flatfile = TRUE,
-                        projection = "grch37"){
-
-  db = config::get("database_name")
```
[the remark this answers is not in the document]
mattssca: Yes, I thought this is something we wanted, to move away from the database. If not, I can easily add this back in 😄

R/database.R
```diff
@@ -1278,81 +1356,101 @@ get_manta_sv = function(min_vaf = 0.1,
...
+  
   if(!missing(region) || !missing(chromosome)){
     suppressWarnings({
       if(grepl("chr",chromosome)){
         chromosome = gsub("chr", "", chromosome)
       }
     })
+    
     all_sv = all_sv %>%
-      dplyr::filter((CHROM_A == chromosome & START_A >= qstart & START_A <= qend) | (CHROM_B == chromosome & START_B >= qstart & START_B <= qend)) %>%
```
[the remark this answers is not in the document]
mattssca: The updates to this function are a combination of improving readability, as well s adding support for specifying multiple sample IDs as a vector. This is related to the issues you brought up in [this](https://morinlabsfu.slack.com/archives/C0224H120CU/p1674625364603939) Slack post. Hope it makes sense.

R/database.R
```diff
@@ -1704,7 +1820,7 @@ get_sample_cn_segments = function(this_sample_id,
     all_segs = all_segs %>%
       dplyr::mutate(chrom = gsub("chr", "", chrom))
   }else{
-    if(!grepl("chr", all_segs$chrom)){
+    if(!grepl("chr", all_segs$chrom[1])){
```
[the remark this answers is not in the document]
mattssca: This is a good point, for now, I'll address this in an update. Thanks!

R/utilities.R
```diff
@@ -35,18 +35,29 @@ compare_coding_mutation_pattern = function(maf_df1,maf_df2,gene){
...
 #'
 #' @examples
-write_sample_set_hash = function(update=TRUE,new_sample_sets_df){
-  sample_sets_file = paste0(config::get("repo_base"),"data/metadata/level3_samples_subsets.tsv")
-  md5_file = paste0(config::get("repo_base"),"data/metadata/level3_samples_subsets_hashes.tsv")
+#' 
+write_sample_set_hash = function(update = TRUE,
+                                 new_sample_sets_df){
+
+  sample_sets_file = paste0(check_config_value(config::get("repo_base")),"data/metadata/level3_samples_subsets.tsv")
```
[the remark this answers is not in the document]
mattssca: Yeah, I think that's a good idea. I'll make the necessary updates. Thanks!

NAMESPACE
rdmorin: You could always update the code to say `stats::p.adjust()`	when it's called or add stats to the @imports line, right?

R/utilities.R
mattssca: I'll go over all the code that currently imports `utils` based on the information you provided. In most cases, the functions called from this package are `head`, `tail`, `read_table` and `write_table`.

R/viz.R
mattssca: Good catch, this is not intended.

mattssca:
Thanks for your comments @Kdreval, I'll work on an update addressing the issues you raised. Stay tuned for a re-review!

NAMESPACE
mattssca: Ok, thanks for clarifying! Makes sense.

mattssca: FYI, after educating myself on what packages are default R packages (i.e loaded every time you start R), I found two additional packages that can be removed for the same reasons as `utils` and `stats`. These packages are `parallel` and `grid`.  I have removed them from being imported by any functions (i/e preventing them to populate NAMESPACE).

Kdreval: Thanks for checking!

R/annotation.R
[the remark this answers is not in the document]
Kdreval: I was wondering about this one here. It seems that the way example is set up it already loads the maf file that was generated as an output of this function. If you try this on the grande maf that is already bundled, does it annotates any variants to be deblacklisted? If so, we can use that to re-work the example here

R/database.R
```diff
@@ -1279,81 +1356,101 @@ get_manta_sv = function(min_vaf = 0.1,
     all_sv = bind_rows(all_sv, missing_sv)
     
   }else{
-    con = DBI::dbConnect(RMariaDB::MariaDB(), dbname = db)
-    all_sv = dplyr::tbl(con, table_name) %>%
-      as.data.frame()
+    stop("database usage is deprecated, please set from_flatfile to TRUE...")
```
Kdreval, line 1359: See this now

R/database.R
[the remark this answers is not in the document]
Kdreval: I missed here that the other filtering steps were taken out of the if statement. Can you please confirm that the output after this modification is identical to what is currently returned in a version that is on master?

R/checks.R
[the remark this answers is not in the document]
rdmorin: Wow. What does this line actually accomplish? I'm surprised R doesn't complain about this syntax.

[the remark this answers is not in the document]
mattssca: This line creates two variables (local_base and base_path) pointing to the same path ("/projects/nhl_meta_analysis_scratch/gambl/results_local/"). 

Yeah, it's kinda strange that R accepts this. 

From the looks of it, this double assignment was added 7 months ago by Ryan. 

I will drop the `base_path` part since it, from the looks of it, doesn't do anything in the function. 

Sounds good?

R/annotation.R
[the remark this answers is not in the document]
rdmorin: My understanding is that Adam is now using a spellchecking add-on for VS Code. Maybe we should all try to use that if we're writing documentation?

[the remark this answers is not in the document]
Kdreval: Yes! Let's try to adopt it! Adam can you advertise this extension on Slack so we all will use the same add on (and any settings that need to be modified from default)? Thanks!

R/checks.R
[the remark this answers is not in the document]
rdmorin: Yes it must be a pasting error. If you see something like this with weird syntax again please consider it a likely bug and fix it if you can.

R/annotation.R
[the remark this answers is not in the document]
mattssca: Yes, It's working out great. The extension I am using is called Spell Right. I'll advertise it on Slack!

R/annotation.R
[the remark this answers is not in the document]
mattssca: This example has been updated to use the bundled grande MAF data (has not been deblacklsited).

Kdreval (approved):

rdmorin (approved):

Issue #166, opened 2023-02-28
prettyForestplot is broken


Kdreval:
I found that my recent update to the prettyForestplot that removed one of the dependencies and added functionality to report mutated/WT counts for each comparison group also introduced bugs. This is how the example looks now :point_down: :thumbsdown: 

![image](https://user-images.githubusercontent.com/47579823/221762414-7b4d181c-0e0f-4d28-873a-ec7e7cf97fb1.png)

The barplot on the right is always for one "gene" showing the same values and the forest on the right is all over the place with broken ordering.
This needs a fix.

Kdreval:
Was fixed in the PR #167

PR #167, opened 2023-02-28
bug fix: prettyForestplot bugs

Kdreval:
I have recently broken the `prettyForestplot`. This PR fixes the bugs described in #166

### DIFF
--- a/R/viz.R
+++ b/R/viz.R
@@ -1427,7 +1427,7 @@ prettyOncoplot = function(maftools_obj,
       warning("mintMutationPercent option is not available when you provide your own oncomatrix. Feel free to implement this if you need it")
       return()
     }
-    mutation_counts <- maftools_obj@gene.summary %>% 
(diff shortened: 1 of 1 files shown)


mattssca (approved):
Thanks, Kostia! I will approve this PR with a minor comment that you can deal with as you like.  Since `dplyr` is specified under import, and because the `mutate` function does not have any conflicting function names from any other package, it is not necessary to specify the package before calling this function (i.e `mutate::dplyr`). It can simply just be `mutate`. Thanks!

Kdreval:
> Thanks, Kostia! I will approve this PR with a minor comment that you can deal with as you like. Since `dplyr` is specified under import, and because the `mutate` function does not have any conflicting function names from any other package, it is not necessary to specify the package before calling this function (i.e `mutate::dplyr`). It can simply just be `mutate`. Thanks!

Oh good! Thanks for the update!

Issue #181, opened 2023-03-15
Set get_coding_ssm_status priorities straight


Kdreval:
The way it is currently set up, the output of `get_coding_ssm_status` will always make hotspots non-redundant with SSM even in the samples that are actually multihit and contain both hot spot and non hot spot mutation. This is not always desirable and should be dictated by user.

This function needs an optional argument that will allow user to dynamically keep multihit samples annotated for both SSM and hot spots.

Kdreval:
This is now fixed in the #182

PR #182, opened 2023-03-15
Fix for `get_coding_ssm_status` function

Kdreval:
This PR is a fix for the issue #181 . Minimal reproducible example for the new functionality:

```

library(tidyverse)
library(vroom)

setwd("~/GAMBLR/")

devtools::load_all()

# get meta
meta <- get_gambl_metadata(
    case_set = "DLBCL-unembargoed"
)

# get mutations
maf <- get_coding_ssm(
        these_samples_metadata = meta
    ) %>%
    dplyr::filter(
        Hugo_Symbol == "CREBBP" # subset to example gene
    )

new_way <- get_coding_ssm_status(
        gene_symbols = "CREBBP",
        maf_data = maf,
        these_samples_metadata = meta
    )

new_way_multihit <- get_coding_ssm_status(
        gene_symbols = "CREBBP",
        maf_data = maf,
        keep_multihit_hotspot = TRUE,
        these_samples_metadata = meta
    )

setdiff(new_way_multihit, new_way)
```
This will show difference in 7 samples where the `CREBBP` is actually multihit and therefore both hot spot and non hot spot should be annotated.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -619,6 +619,7 @@ get_coding_ssm_status = function(gene_symbols,
                                  maf_path = NULL,
                                  maf_data,
                                  include_hotspots = TRUE,
+                                 keep_multihit_hotspot = FALSE,
                                  recurrence_min = 5,
                                  seq_type = "genome",
                                  projection = "grch37",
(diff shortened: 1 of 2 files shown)


Kdreval:
This has been updated to not only report non-hotspot mutations when the user asks to keep multihit samples, but also to keep track of how many non-hotspot mutations are in there. This is also documented in the argument description.

rdmorin (approved):

PR #191, opened 2023-04-12
Add IGV functionality

rdmorin:
# Pull Request Checklists

### Required

### Required

Example:
```
#'
```

Example:
```
#' @return nothing
#' @export
#' @import tidyverse ggrepel
```


R/utilities.R
```diff
@@ -3015,6 +3020,116 @@ get_bams = function(sample,
   return(bam_details)
 }
 
+#' @title View a variant in IGV
+#'
+#' @description Load bam(s) and view the context around a mutation
+#'
+#' @details 
+#'
+#' @param this_mutation Specify the mutation of interest in MAF format.
```
mattssca, line 3029: Multiple parameter descriptions seem to be missing.

R/utilities.R
```diff
@@ -3015,6 +3020,116 @@ get_bams = function(sample,
...
+#' @param igv_port Specify the port IGV is listening on.
+#' @param socket 
+#' @param sort_by base, quality, sample or readGroup
+#'
+#' @return Path to file (.png).
+#'
+#' @import SRAdb
+#' @export
+#'
+#' @examples
```
mattssca, line 3039: I think this example is a good candidate for an example that should not be triggered with `build_site`. Consider wrapping the examples in `\dontrun{}`.

R/utilities.R
```diff
@@ -3048,29 +3163,120 @@ get_bams = function(sample,
...
+#' # make_igv_snapshot(socket=socket,sample_ids=this_mutation$Tumor_Sample_Barcode,this_seq_type="capture",colour_by="READ_STRAND")
+#' # run on a bunch of variants using apply:
+#' # apply(to_snapshot,1,function(x){make_igv_snapshot(sample_ids=x["sample_id"],
+#'       seq_type_filter="capture",chrom=x["chr"],start=x["start"],end=x["end"],
+#'       details=paste0(x["ref"],"-",x["alt"]),
+#'       gene=x["Hugo_Symbol"],socket=socket)})
+#'       
 make_igv_snapshot = function(bams,
+                             sample_ids,
+                             this_seq_type="genome",
```
mattssca, line 3179: Update the parameter descriptions to match the new/deprecated parameters of this function.

R/utilities.R
```diff
@@ -3015,6 +3020,116 @@ get_bams = function(sample,
...
+#' @details 
+#'
+#' @param this_mutation Specify the mutation of interest in MAF format.
+#' @param igv_port Specify the port IGV is listening on.
+#' @param socket 
+#' @param sort_by base, quality, sample or readGroup
+#'
+#' @return Path to file (.png).
+#'
+#' @import SRAdb
```
mattssca, line 3036: What functions do we use from this package (SRAdb)? Do we want to load the package or just call the function from this package explicitly (SRAdb::function)?

mattssca (changes requested):
Thanks for this update, Ryan. I have some comments related to the function documentation. As a side note, make sure to run `devtools::document()` to update the affected `man/*.Rd`. There are also merge conflicts between this branch and the master that should be resolved. Also, the Git Action did fail for this PR. Thanks!

R/utilities.R
```diff
@@ -3015,6 +3020,116 @@ get_bams = function(sample,
   return(bam_details)
 }
 
+#' @title View a variant in IGV
+#'
+#' @description Load bam(s) and view the context around a mutation
+#'
+#' @details
```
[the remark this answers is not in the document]
rdmorin: I've moved them as suggested. Not sure how the formatting of code will look when it's in the details section though (or if theres a good way to ensure it's formatted nicely). Do you know?

R/utilities.R
rdmorin: I think this is resolved now

R/utilities.R
rdmorin: Good call. Fixed.

R/utilities.R
rdmorin: I hope most are self-explanatory. I tried to fix the ones with issues.

R/utilities.R
rdmorin: Everything IGV related needs this package. I'm not sure if we need to load it but feel free to test it if you want. testers of these new functions would be greatly appreciated. @vladimirsouza may also want to try this

R/utilities.R
```diff
@@ -3094,7 +3094,8 @@ get_gambl_colours = function(classification = "all",
 get_bams = function(this_sample_id,
                     this_patient_id){
 
-  meta = get_gambl_metadata(tissue_status_filter = c("tumour", "normal"), seq_type_filter = "genome")
+  meta = get_gambl_metadata(tissue_status_filter = c("tumour", "normal"), seq_type_filter = seq_type_filter)
```
Kdreval, line 3097: I think here and in the next line the `seq_type_filter` is not defined for this function, we probably need to add it as the function parameter

R/utilities.R
```diff
@@ -3115,29 +3117,34 @@ get_bams = function(this_sample_id,
   }else{
     igv_build = build
   }
-  tumour_genome_bams = dplyr::filter(meta_patient, seq_type == "genome" & tissue_status == "tumour") %>%
-    dplyr::pull(data_path)
+  bam_path_pattern = "/projects/rmorin/projects/gambl-repos/gambl-rmorin/data/{seq_type}_bams/{sample_id}.{genome_build}.bam"
```
Kdreval, line 3120: Should we have this pulled from the config?  Like `config::get("repo_path")` and paste with the data/... to generate the string with path, so it is consistent with the rest of the functions?

R/utilities.R
```diff
@@ -3146,6 +3153,116 @@ get_bams = function(this_sample_id,
...
+  chrom = pull(this_mutation,Chromosome)
+  region = paste0(chrom,":",start-50,"-",end+50)
+  sock = socket
+  
+    meta = get_gambl_metadata(seq_type_filter=this_seq_type) %>% 
+      dplyr::filter(sample_id %in% this_sample_id)
+    
+      genome_build = pull(meta,genome_build)
+    
+    bam_path_pattern = "/projects/rmorin/projects/gambl-repos/gambl-rmorin/data/{seq_type}_bams/{sample_id}.{genome_build}.bam"
```
Kdreval, line 3214: Maybe this piece here can be swapped to use the `get_bams()` output? But this can be done in another PR with small improvements

Kdreval (approved):
Some small comments, I think the main one is about defining the `seq_type_filter` parameter for `get_bams()`

R/utilities.R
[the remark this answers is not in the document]
mattssca: I think you can wrap the code under the details section in this way:
```
#' @title My Function
#' @description Description of the function
#' @details Here are some details about the function
#'
#' ```
#' code block goes here
#' ```
#'
#' @param
#' @examples
#' @export
#' @import
```

R/utilities.R
rdmorin: I think the problem here is I couldn't tell what function you were commenting on because the line numbering shown in this interface didn't match the line number in my file and the code context I see is not allowing me to find the function. What function are you commenting on?

mattssca (approved):

Issue #213, opened 2023-06-12
`gene_to_region` function changes input order of genes


vladimirsouza:
`gene_to_region` always sorts the order of the output regions by chromosome (but incorrectly; `chromosome` is `character`) and start location. In this way, it's impossible to know to which gene each output region refers. An example:
```
> my_genes <- c("MYC", "BCL2", "imaginary_gene")
> gene_to_region(gene_symbol = my_genes)
# 2 region(s) returned for 2 gene(s)
# [1] "18:60790579-60987361"  "8:128747680-128753674"
```
In this example, The first region `"18:60790579-60987361"` refers to `"BCL2"` (second input gene), and the second region `"8:128747680-128753674"` refers to `"MYC"` (first gene). `"imaginary_gene"` was ignored.

My suggestion is that the order of the output regions follows the same order as the input genes (they should have the same lengths). Besides that, the function could output a named vector to inform the gene names as well. An example:
```
> my_genes <- c("MYC", "BCL2", "imaginary_gene")
> gene_to_region(gene_symbol = my_genes)
#                     MYC                    BCL2          imaginary_gene 
# "8:128747680-128753674"  "18:60790579-60987361"                      NA 
```

If useful, we can add a new boolean parameter `sort_regions`, where `TRUE` (must be specified by the user; default is FALSE) means to sort regions by their chromosome and start location (`NA`s always to the right hand).
```
> my_genes <- c("MYC", "imaginary_gene", "BCL2")
> gene_to_region(gene_symbol = my_genes, sort_regions = TRUE)
#                     MYC                    BCL2          imaginary_gene 
# "8:128747680-128753674"  "18:60790579-60987361"                      NA 
```

If anyone finds this change useful, I would be happy in making the change and submit a Pull Request.

mattssca:
Thanks for posting this Vladimir. I like your suggestion that the returned regions follow the same order as the genes given to the function. However, have you played around with the `return_as` parameter? If set to "df" or "bed", the returned data frame will tell you what genes correspond to what regions.

```
> my_genes <- c("MYC", "BCL2", "imaginary_gene")
> gene_to_region(gene_symbol = my_genes, return_as = "df")
2 region(s) returned for 3 gene(s)
  chromosome     start       end gene_name hugo_symbol ensembl_gene_id
1         18  60790579  60987361      BCL2        BCL2 ENSG00000171791
2          8 128747680 128753674       MYC         MYC ENSG00000136997
```

[...]

Issue #208, opened 2023-06-06
New wrapper function to produce the SSM and CN joint matrix


Kdreval:
We have functionality in GAMBLR to generate the feature matrix for simple mutations (`get_coding_ssm_status()`) and another function that returns copy number state of the gene of interest (`get_cn_states()`). What is missing is the wrapper function that will run both of these to generate a single binary matrix (0 for no feature and 1 for the presence of the feature) where either mutation or CNV will be considered.

Here is an example:
```
my_genes <- "MYC"

my_regions <- grch37_lymphoma_genes_bed %>%
	filter(hgnc_symbol %in% my_genes) %>%
	mutate(region = paste0(
		chromosome_name,
		":",
		start_position,
		"-",
		end_position
	)) %>%
	pull(region)

example_ids <- c(
	"BLGSP-71-06-00160-01A-03D",
	"BLGSP-71-06-00252-01A-01D",
	"BLGSP-71-19-00122-09A.1-01D",
	"BLGSP-71-19-00523-09A-01D",
	"BLGSP-71-21-00187-01A-01E",
	"BLGSP-71-21-00188-01A-04E"
)


my_meta <- get_gambl_metadata() %>%
	filter(sample_id %in% example_ids)

my_maf <- get_ssm_by_samples(
	these_samples_metadata = my_meta
)

cn_matrix <- get_cn_states(
	regions_list = my_regions,
	these_samples_metadata = my_meta,
	region_names = my_genes
)

ssm_matrix <- get_coding_ssm_status(
	gene_symbols = my_genes,
	these_samples_metadata = my_meta,
	maf_data = my_maf,
	from_flatfile = FALSE,
	include_hotspots = FALSE
)
```

In this example, 3 samples do not have SSM (the ssm_matrix has 0 for the mutation presence), but they have CNV (the cn_matrix has copy number higher than 2). The new function will aggregate these events, and all samples in the example will have 1 for the combined feature MYC_Mut_or_AMP

```
                    sample_id MYC_Mut_or_AMP
1   BLGSP-71-06-00160-01A-03D              1
2   BLGSP-71-06-00252-01A-01D              1
3 BLGSP-71-19-00122-09A.1-01D              1
4   BLGSP-71-19-00523-09A-01D              1
5   BLGSP-71-21-00187-01A-01E              1
6   BLGSP-71-21-00188-01A-04E              1
```

[...]

PR #218, opened 2023-06-14
Change `gene_to_region` and add `get_cnv_and_ssm_status`

vladimirsouza:
This pull request is a response to the [#213](https://github.com/morinlab/GAMBLR/issues/213) and [#208](https://github.com/morinlab/GAMBLR/issues/208) issues.

### About [#213](https://github.com/morinlab/GAMBLR/issues/213) issue (gene_to_region function changes input order of genes):

- Correct how to sort regions by chromosome and start location.
- If `return_as = "region"`, return a named vector where names are gene ids.
- Add boolean parameter `sort_regions` (default is TRUE). If FALSE, the order of the output regions follows the same order of the input genes.
- Remove the boolean parameter `na_for_genes_not_found`. The function never outputs NAs for genes not found anymore. If there are genes not found, the function prints a list with these genes.
- Possible values for `genome_build` argument are `grch37` or `hg38`.

Please, also see my last (June 26) comment on each file. 

### About [#208](https://github.com/morinlab/GAMBLR/issues/208) issue (New wrapper function to produce the SSM and CN joint matrix):

- `get_cnv_and_ssm_status` uses updated `gene_to_region`.
- My description of `augmented` argument is different from the ones in `get_ssm_by_samples`'s and `get_coding_ssm_status`'s help (I find them confusing). Please, check if it keeps the same meaning. My description: ```Argument to internally pass to the functions `get_ssm_by_samples` and `get_coding_ssm_status`. A logical parameter (default: TRUE). Set to FALSE to use multi-sample patients, instead of the original MAF from each sample.```. Description in get_ssm_by_samples and get_coding_ssm_status: `default: TRUE. Set to FALSE if you instead want the original MAF from each sample for multi-sample patients instead of the augmented MAF.`
- I added the arguments `subset_from_merge` and `augmented` because if using their default values, the `min_read_support_ssm` argument (it's called `min_read_support` in `get_ssm_by_samples` and `get_coding_ssm_status`) is useless. When running `get_ssm_by_samples` with default values for these arguments, the `get_ssm_by_sample` (without 's') function is called internally, which in turn doesn't use `min_read_support` because the `/projects/nhl_meta_analysis_scratch/gambl/results_local/gambl/slms_3-1.0_vcf2maf-1.3/99-outputs/deblacklisted/augmented_maf/genome--projection/BLGSP-71-06-00160-01A-03D--BLGSP-71-06-00160-12A-01D--matched.slms-3.final.grch37.maf` file doesn't exists (see [code](https://github.com/morinlab/GAMBLR/blob/master/R/database.R#L525)). P.S.: If we don't give both arguments `maf_data` and `maf_path` to `get_coding_ssm_status`, it calls `get_coding_ssm` with fixed value for `min_read_support` = 3 ([code](https://github.com/morinlab/GAMBLR/blob/master/R/utilities.R#L671)).
- I figured out why there might be NAs in the output and this is not an error in the function. NAs happen while running `get_cn_states` and the reason is that `get_cn_segments` doesn't return CN segments for some combinations of genes and samples. Here is an example:
- I figured out why there might be NAs in `get_cnv_and_ssm_status`'s output, and this is not an error of the function. NAs happen while running `get_cn_states` and the reason is that `get_cn_segments` doesn't return CN segments for some combinations of genes and samples. Here is an example:
```
# input data
genes_and_cn_threshs <- data.frame(gene_id=c("MYC", "MIR17HG"), cn_thresh=c(3, 3))
these_sample_ids <- c("BLGSP-71-19-00123-09A.1-01D")
seq_type <- "genome"
these_samples_metadata <- get_gambl_metadata(seq_type_filter = seq_type) %>%
  filter(sample_id %in% these_sample_ids)

# get_cnv_and_ssm_status returns NA
get_cnv_and_ssm_status(
  genes_and_cn_threshs = genes_and_cn_threshs,
  these_samples_metadata = these_samples_metadata,
  seq_type = seq_type,
  only_cnv = "all"
)
# 2 region(s) returned for 2 gene(s)
#                             MYC MIR17HG                                                                                              
# BLGSP-71-19-00123-09A.1-01D  NA       0

# for this specified sample, get_cn_segments doesn't return cn segments from MYC
get_cn_segments(region = "8:128747680-128753674",
                streamlined = TRUE,
                this_seq_type = seq_type) %>% 
  filter(ID %in% these_samples_metadata$sample_id)
# [1] ID CN                                                                                                                            
# <0 rows> (or 0-length row.names)

# but get_cn_segments returns cn segments from MIR17HG
get_cn_segments(region = "13:92000074-92006833",
                streamlined = TRUE,
                this_seq_type = seq_type) %>% 
  filter(ID %in% these_samples_metadata$sample_id)
#                            ID CN                                                                                                     
# 1 BLGSP-71-19-00123-09A.1-01D  2
```

# Pull Request Checklists

### Required

```
# I tried many different combinations of arguments. Here are two of them.
gene_to_region(
  gene_symbol = c("BCL2","imaginary_gene", "MYC"),
  # ensembl_id = c("ENSG00000171791","imaginary_gene", "ENSG00000136997"),
  genome_build = c("grch37", "grch38")      [1],
  return_as = c("region", "bed", "df")    [2],
  sort_regions = c(TRUE, FALSE)           [1],
  na_for_genes_not_found = c(TRUE, FALSE) [2]
)
# 2 region(s) returned for 3 gene(s)
#   chromosome     start       end hugo_symbol
# 1          8 128747680 128753674         MYC
# 2         18  60790579  60987361        BCL2

gene_to_region(
  # gene_symbol = c("BCL2","imaginary_gene", "MYC"),
  ensembl_id = c("ENSG00000171791","imaginary_gene", "ENSG00000136997"),
  genome_build = c("grch37", "grch38")      [2],
  return_as = c("region", "bed", "df")    [1],
  sort_regions = c(TRUE, FALSE)           [2],
  na_for_genes_not_found = c(TRUE, FALSE) [1]
)
# 2 region(s) returned for 3 gene(s)
#            ENSG00000171791             imaginary_gene            ENSG00000136997 
#  "chr18:63123345-63320128"                         NA "chr8:127735433-127742951"
```

### Required

- [x] I documented my function using [Roxygen style](https://jozef.io/r102-addin-roxytags/#:~:text=Inserting%20a%20skeleton%20%2D%20Do%20this,Shift%2BAlt%2BR%20).)

- [x] Adequate function documentation (see [new-function documentation template](https://github.com/morinlab/GAMBLR#title) for more info)

- [x] I have ran `devtools::document()` to add the newly created function to NAMESPACE (do not manually add anything to this file!).

- [x] I tested the new code for compatibility with existing functionality in the Master branch (please provide a reprex of how you tested the original functionality)
```
fancy_ideogram(this_sample_id = "HTMCP-01-06-00422-01A-01D",
               gene_annotation = c("BCL2", "imaginary_gene", "MYC"),
               plot_title = "Sample-level Ideogram Example",
               plot_subtitle = "grch37")
```
`fancy_ideogram` uses `gene_to_region` internally. It makes this nice plot [here](https://drive.google.com/file/d/1gjrEXiW7zsVZbMHeziBxD9jzfY6gegnO/view?usp=sharing).


R/utilities.R
```diff
@@ -361,60 +363,90 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+                          na_for_genes_not_found = FALSE){
+  
+  stopifnot('`genome_build` parameter must be "grch37" or "grch38"' = genome_build %in% c("grch37", "grch38"))
+  stopifnot('`return_as` parameter must be "region", "bed" or "df"' = return_as %in% c("region", "bed", "df"))
+  
   #set mart based on selected genome projection
   if(genome_build == "grch37"){
     gene_coordinates = grch37_gene_coordinates
     chr_select = paste0(c(c(1:22),"X","Y"))
-  }else if(genome_build == "hg38"){
```
rdmorin, line 370: This trusts the user to only ever specify one of these two options. Would be better to leave the original check for hg38 and have a trailing else here to catch the case when a user specifies an option that isn't handled.

R/utilities.R
```diff
@@ -361,60 +363,90 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+  }else{
     gene_coordinates = hg38_gene_coordinates
     chr_select = paste0("chr", c(c(1:22),"X","Y"))
   }
-
+  
   #filter on gene_symbol/ensembl_id
   if(!missing(gene_symbol) && missing(ensembl_id)){
-    gene_coordinates = dplyr::filter(gene_coordinates, hugo_symbol %in% gene_symbol)
+    gene_coordinates = gene_coordinates[ match(gene_symbol, gene_coordinates$hugo_symbol), ]
```
rdmorin, line 384: What is the benefit/purpose of these changes?

R/utilities.R
```diff
@@ -361,60 +363,90 @@ get_ssh_session = function(host="gphost01.bcgsc.ca"){
...
+  
   region = dplyr::select(gene_coordinates, chromosome, start, end, gene_name, hugo_symbol, ensembl_gene_id) %>%
-    as.data.frame() %>%
-    dplyr::arrange(chromosome, start) %>%
-    dplyr::filter(chromosome %in% chr_select)
+    as.data.frame()
+  
+  if(sort_regions){
+    if(genome_build == "grch37"){
+      chrm_num = format( region$chromosome, justify = "right", na.encode = FALSE )
```
rdmorin, line 398: Can you explain what the rationale is for these format lines? I don't know what this solves

rdmorin (changes requested):

R/utilities.R
Kdreval: I think this change is for the correct ordering of the chromosomes (1,2,3 instead of 1,11,12..). But this format seems to add a prepending space at the front:

```
r$> chr <- c(rep(1, 5), rep(11, 3), rep("X", 5), rep(2, "5"))

r$> format(chr, justify = "right", na.encode = FALSE )
 [1] " 1" " 1" " 1" " 1" " 1" "11" "11" "11" " X" " X" " X" " X" " X" " 2" " 2" " 2" " 2" " 2"
```

I would  think we would avoid at all costs adding new dependencies because of how many there are already but another way to sort in the correct order is the 
```
r$> as.character(
        sort(
            factor(
                chr,
                levels = c((1:22), "X"),
                ordered = TRUE
            )
       )
    )
 [1] "1"  "1"  "1"  "1"  "1"  "2"  "2"  "2"  "2"  "2"  "11" "11" "11" "X"  "X"  "X"  "X"  "X" 
```

This will also work for the prefixed chromosomes

R/utilities.R
vladimirsouza: I think we should use either `grch37/grch38` or `hg19/hg38`. Using this mix of nomenclature like `grch37/hg38` is a bit weird for me. Please, let me know what you think.

The line 
```
stopifnot('`genome_build` parameter must be "grch37" or "grch38"' = genome_build %in% c("grch37", "grch38"))
```
makes sure that the user doesn't enter a wrong value.

Besides that, I think we always should mention what are the possible values for an argument in its documentation. Here is an example I think we should correct,
```
#' @param genome_build Reference genome build for the coordinates in the MAF file. The default is grch37 genome build.
```

R/utilities.R
vladimirsouza: This change is to ensure that the gene order remains the same as in the input `gene_symbol`. Take a look at this example:
```
gene_symbol <- c("MYC", "STAT6", "CCND3")
gene_coordinates = grch37_gene_coordinates

gene_coordinates[ match(gene_symbol, gene_coordinates$hugo_symbol), ]
# # A tibble: 3 × 6
#   ensembl_gene_id chromosome     start       end gene_name hugo_symbol
#   <chr>           <chr>          <dbl>     <dbl> <chr>     <chr>      
# 1 ENSG00000136997 8          128747680 128753674 MYC       MYC        
# 2 ENSG00000166888 12          57489191  57525922 STAT6     STAT6      
# 3 ENSG00000112576 6           41902671  42018095 CCND3     CCND3      

dplyr::filter(gene_coordinates, hugo_symbol %in% gene_symbol)
# # A tibble: 3 × 6
#   ensembl_gene_id chromosome     start       end gene_name hugo_symbol
#   <chr>           <chr>          <dbl>     <dbl> <chr>     <chr>      
# 1 ENSG00000166888 12          57489191  57525922 STAT6     STAT6      
# 2 ENSG00000112576 6           41902671  42018095 CCND3     CCND3      
# 3 ENSG00000136997 8          128747680 128753674 MYC       MYC        
```

R/utilities.R
vladimirsouza: @Kdreval, you are right, this is to sort the chromosomes. `format` function does add spaces to the strings, but this is not a problem here, because `format` output goes to a separate object, called `chrm_num`, that I use just for the sorting. The real chromosome values are unchanged in `region$chromosome`.

vladimirsouza: `format` function is from the `base` packege.

R/utilities.R
vladimirsouza: The next line "assigning the hygo_symbol column value of the gene_symbol" is to deal with samples that don't exist. Then we don't miss the name of missing genes. Example:
```
gene_symbol <- c("MYC", "STAT6", "imaginary_gene", "CCND3")
gene_coordinates = grch37_gene_coordinates
gene_coordinates = gene_coordinates[ match(gene_symbol, gene_coordinates$hugo_symbol), ]
gene_coordinates
# A tibble: 4 × 6
#   ensembl_gene_id chromosome     start       end gene_name hugo_symbol
#   <chr>           <chr>          <dbl>     <dbl> <chr>     <chr>      
# 1 ENSG00000136997 8          128747680 128753674 MYC       MYC        
# 2 ENSG00000166888 12          57489191  57525922 STAT6     STAT6      
# 3 NA              NA                NA        NA NA        NA         
# 4 ENSG00000112576 6           41902671  42018095 CCND3     CCND3      

gene_coordinates$hugo_symbol <- gene_symbol
gene_coordinates
# # A tibble: 4 × 6
#   ensembl_gene_id chromosome     start       end gene_name hugo_symbol   
#   <chr>           <chr>          <dbl>     <dbl> <chr>     <chr>         
# 1 ENSG00000136997 8          128747680 128753674 MYC       MYC           
# 2 ENSG00000166888 12          57489191  57525922 STAT6     STAT6         
# 3 NA              NA                NA        NA NA        imaginary_gene
# 4 ENSG00000112576 6           41902671  42018095 CCND3     CCND3         
```

R/utilities.R
Kdreval: I agree sticking to the same system would make a lot of sense and simplify significantly the amount of time and effort needed to handle the chromosome prefixes, but that's not how it happened historically and most of the tools either support non-prefixed data in grch37 projection or chr-prefixed in hg38, so this convention is in the core of GAMBLR (and other projects) as well. This way, the data generated/obtained here is directly compatible with tools like vcf2maf, mutsig, oncodrive etc. So I think it is good to keep here the grch37/hg38 format for consistency and modify this stopifnot to catch the cases that aren't handled.

R/utilities.R
rdmorin: why would we want those back with all NA? I guess this is meant to alert the user that something went wrong?

Kdreval: Thanks for the example, it is clear now. I think this might be addressed by piping this to the `dplyr::arrange()` to match the specified gene order, so we're sticking with dplyr syntax here for consistency

R/utilities.R
rdmorin: Historically, our two main genome builds in GAMBL are hg38 and grch37. Accordingly, our projections have been named that way since a projection is really just a result that has been "converted" to a new coordinate system. We need to stick with hg38 and grch37 even if it's confusing. There are other nuances such as the fact that hg38 has chr-prefixed coordinates and grch37 doesn't and we try to keep things consistent with the underlying genome build associated with the projection.

R/utilities.R
Kdreval: Does it still do a proper match later when the chromosomes are ordered? Just not sure how the comparison of ` 1` to `1` is made to ensure the correct order.

rdmorin: Oh I see. I had missed this line earlier. 

```region = dplyr::arrange(region, chrm_num, start)```

@Kdreval shouldn't we be only doing this chr prefix removal for the grch37 case (i.e. in the else/hg38 context it should be left prefixed)?

File: R/utilities.R (hunk 1 above)
Comment (vladimirsouza) [reply R/utilities.R:370]: I think there is confusion regarding what I said. If I understand correctly, `grch37/grch38` are two versions of the human genome assembly and these are the nomenclature used by NCBI. `hg19/hg38` are technically the "same" versions but are the nomenclature used by UCSC.
Just to be clear, I'm not saying that we should always use the same assembly version, but the same nomenclature. 
grch37 = hg19 and grch38 = hg38.

File: R/utilities.R (hunk 2 above)
Comment (rdmorin) [reply R/utilities.R:384]: I think he's using data.table match here to ensure genes not in the data frame get added but I'm not sure this is really a feature we want. Might be easier just to warn the user about any genes that had no region info available.

Comment (Kdreval) [reply R/utilities.R:384]: I agree this can be thrown as a warning to the user and does not need to be present in the output as NA.

File: R/utilities.R (hunk 3 above)
Comment (Kdreval) [reply R/utilities.R:398]: Right, this only does the `chr` handling for the else part of `if(genome_build == "grch37")` check, so I think this alright - but the order handling is done for both projections because when the `X` is present they are all treated as characters anyway and will result in that non-sensible 1,11,12..2... order

File: R/utilities.R (hunk 1 above)
Comment (rdmorin) [reply R/utilities.R:370]: It's more nuanced than that though. hg19 is not the same as grch37. grch38 is not the same as hg38. Our two projections are:
grch37 (not hg19)
hg38 (not grch38)

The main differences are:
hg19, hg38 - chr-prefixed chromosome names
grch37, grch38 - bare chromosome names (do not start with chr prefix)
*the mitochondrial chromosome and other chromosomes may also have distinctions. I can't recall off the top of my head

Comment (vladimirsouza) [reply R/utilities.R:370]: Yes, they are not exactly the same. That is the reason why I wrote that `grch37/grch38` and `hg19/hg38` are 
```
technically the "same" versions
```
with `same` in quotes. 

But now I understood. We really have `grch37` and `hg38` and it's not only the nomenclature. I totally agree with your change.

Comment (rdmorin) [reply R/utilities.R:370]: Exactly. If a user requests anything other than grch37 or hg38 we should throw an error for them since we don't support any other reference versions.

File: R/utilities.R (hunk 3 above)
Comment (rdmorin) [reply R/utilities.R:398]: I'm wondering out loud if this code and other code normalizing coordinates to builds could eventually shift into Laura's new process_coordinates helper function (still in a PR)

File: R/utilities.R (hunk 2 above)
Comment (vladimirsouza) [reply R/utilities.R:384]: I got this idea of returning NAs (adding the argument `na_for_genes_not_found`) because I wanted the output to have the same length and order as the input gene names. I needed that to write my function to report CNV and SSM combined status. But I change the code and I'm not using this even in my function. Maybe we should remove this `na_for_genes_not_found` parameter any never return NAs.

Comment (Kdreval) [reply R/utilities.R:384]: This sounds good, I agree - and also think we need to implement Ryan's suggestion of giving a warning for any gene that would be NA here, so at least the user is aware of this

File: R/utilities.R (hunk 1 above)
Comment (vladimirsouza) [reply R/utilities.R:370]: I changed the possible values for the `genome_build` argument to `grch37` or `hg38`.

File: R/utilities.R (hunk 2 above)
Comment (vladimirsouza) [reply R/utilities.R]: Changes that I made:
- the argument `na_for_genes_not_found` has been removed and output never includes NAs;
- If genes have no region info available, a message listing these genes is printed, e.g.
```
Some input gene(s) have no region info available. They are:
imaginary_gene.
8 region(s) returned for 9 gene(s)
```
- This part of the code was rewritten with `dplyr` syntax for consistency.

File: R/utilities.R (hunk 3 above)
Comment (vladimirsouza) [reply R/utilities.R]: I'm not using `format` function anymore, but @Kdreval suggestion to use chromosome numbers as ordered factors:
```
chrm_num = factor(chrm_num, levels = c(1:22, "X", "Y"), ordered = TRUE)
```

vladimirsouza:
I've addressed the feedback received in this PR. Please, review this PR.

rdmorin (approved):

PR #230, opened 2023-07-10
Remove remote set-up from vignettes

vladimirsouza:
# Pull Request Checklists

### Required

- [x] I updated and/or successfully knitted a vignette that relies on the modified code (which ones?)

- [x] I tested the new code for compatibility with existing functionality in the Master branch (please provide a reprex of how you tested the original functionality)


R/checks.R
```diff
@@ -169,7 +169,7 @@ check_file_details = function(relative_paths){
   not_found = c()
   base_path = check_config_value(config::get("project_base"))
   for(relative_path in relative_paths){
-    full_path = paste0(base_path,relative_path)
+    full_path = file.path(base_path, relative_path)
```
Kdreval, line 172: Do you happen to have an example of the error this was generating? The `paste0` approach is used extensively to get the path to file and hasn't been failing so far - so wondering to check hot it was behaving here.

Kdreval: Got it, thanks for the example. Will it still work as expected if the user will specify the project_base correctly, with trailing `/` ? Just want to confirm that the presence of `//` when this function is used won't cause any issues downstream

README.html
```diff
@@ -176,7 +176,7 @@ <h2>Installation</h2>
 <p>If you have access to gphost, the easiest way to obtain and run GAMBLR is to do this via Rstudio on a gphost. If you do not have access to gphost, please refer to the <a href="#Run%20Remote%20On%20A%20Local%20Machine">Run Remote On A Local Machine</a> section. Assuming you are running Rstudio on gphost, clone the repo to your home directory (not your GAMBL working directory).</p>
 <pre><code>git clone git@github.com:morinlab/GAMBLR.git</code></pre>
 <p>In Rstudio (on a gphost), set your working directory to the place you just cloned the repo.</p>
-<pre><code>setwd(&quot;~/GAMBLR-master&quot;)</code></pre>
+<pre><code>setwd(&quot;~/GAMBLR&quot;)</code></pre>
```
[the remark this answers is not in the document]
Kdreval: Perfect, thanks

vignettes/cn_sv_tutorial.Rmd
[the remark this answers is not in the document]
Kdreval: My understanding that the vignette is created once you knit it and if there are any changes it will show up as modified and be picked up by git, so you can commit it. If the git status does not show the html files as changed then there are no changes to commit and it can stay as is

R/checks.R
vladimirsouza: Yes, it works. When I run `check_gamblr_config` to check if there are missing files (it's not `check_file_details` which does that, it is called internally and was where I made the change), files with the `//` pattern could be found. For example:
```
Looking for: /home/vladimir/repos/gambl_results//all_the_things/cnv_master-1.0/merges/genome--hg38.seg
OK
```
It's not beautiful, but robust. If needed, I could add the line `gsub("//", "/", filePath)` to let the function always outputs a single `/` (I would go with that). 

Some more examples to show that paths with `//` works:
```
file.exists("/home/vladimir/repos/gambl_results//all_the_things/cnv_master-1.0/merges/genome--hg38.seg")
# [1] TRUE
readLines("/home/vladimir/repos/gambl_results//all_the_things/cnv_master-1.0/merges/genome--hg38.seg", n = 5)
# [1] "ID\tchrom\tstart\tend\tLOH_flag\tlog.ratio"              "KM-H2\tchr1\t10001\t16849162\t0\t0"
# [3] "KM-H2\tchr1\t16849163\t101772836\t0\t0.5849625007211563" "KM-H2\tchr1\t101772837\t109541401\t0\t0"
# [5] "KM-H2\tchr1\t109541402\t119838793\t1\t0"
```

Kdreval: Thanks for confirming, I think it is good then :+1:

Kdreval (approved):

PR #238, opened 2023-07-26
Kdreval fixes

Kdreval:
This PR adds an option for the user to choose engine (read_tsv or grep) to the get_gene_expression functionality. It was found that on the GSC the grep takes 10 min to execute, while read_tsv is 40% faster (6 min). While there is a difference in speed, there is also RAM consideration to take into account, so therefore a new argument `engine` was added to allow user to choose different options.


Kdreval:
This is transferred to PR in GAMBLR.results ([#21](https://github.com/morinlab/GAMBLR.results/pull/21))

